PR #12, opened 2015-01-13
Ignore the submissions that have 0 score.

muzaffaryousaf:
Ignore the submissions that have 0 score.

https://openedx.atlassian.net/browse/TNL-1125

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -377,6 +377,7 @@ def get_top_submissions(course_id, item_id, item_type, number_of_top_scores, use
     """Get a number of top scores for an assessment based on a particular student item
 
     This function will return top scores for the piece of assessment.
+    It will ignore the submissions that have 0 score/points_earned.
     A score is only calculated for a student item if it has completed the workflow for
     a particular assessment module.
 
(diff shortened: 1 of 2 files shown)


submissions/api.py
```diff
@@ -467,8 +469,8 @@ def get_top_submissions(course_id, item_id, item_type, number_of_top_scores, use
             for score in scores
         ]
 
-    # Always store the retrieved list in the cache
-    cache.set(cache_key, top_submissions, TOP_SUBMISSIONS_CACHE_TIMEOUT)
+        # Always store the retrieved list in the cache
+        cache.set(cache_key, top_submissions, TOP_SUBMISSIONS_CACHE_TIMEOUT)
```
muzaffaryousaf: @stephensanchez does this make sense to set the cache only after we hit the database. Please let me know what do you think ? 
Thanks.

muzaffaryousaf:
@stephensanchez can you please review the code for this PR.
Thanks!

submissions/api.py
stephensanchez: It makes sense to me to set it after the database call. We check on line 440 if the top submissions are cached yet, if not, we query, then set the results in the cache. What did you have in mind?

stephensanchez: Oh I think I understand now, you're referring to whether your change is valid. It seems reasonable to me.

stephensanchez:
:+1:

submissions/api.py
muzaffaryousaf: Yeah, Before we were setting the cache in any case either we hit the db or cache. Now, I've changed this and only setting the cache when we hit the db.

stephensanchez: Makes sense!

PR #14, opened 2015-04-25
Add get_all_submissions method to the API

bradenmacdonald:
This adds a new API method to edx-submissions:
```python
get_all_submissions(course_id, item_id, item_type, read_replica=True)
```

This method can be used to fetch all student answers to a particular courseware item, e.g. for CSV export.

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -401,14 +401,16 @@ def get_all_submissions(course_id, item_id, item_type, read_replica=True):
     submission_qs = Submission.objects
     if read_replica:
         submission_qs = _use_read_replica(submission_qs)
+    # We cannot use SELECT DISTINCT ON because it's PostgreSQL only, so unfortunately
+    # our results will contain every entry of each student, not just the most recent.
+    # We sort by student_id and primary key, so the reults will be grouped be grouped by
+    # student, with the most recent submission being the first one in each group.
     query = submission_qs.select_related('student_item').filter(
         student_item__course_id=course_id,
         student_item__item_id=item_id,
         student_item__item_type=item_type,
-    ).order_by('student_item__student_id', '-created_at').iterator()
+    ).order_by('student_item__student_id', '-submitted_at', '-id').iterator()
 
-    # We cannot use SELECT DISTINCT ON because it's PostgreSQL only, so unfortunately
-    # our results will contain every entry of each student, not just the most recent.
     for unused_student_id, row_iter in itertools.groupby(query, operator.attrgetter('student_item.student_id')):
         submission = next(row_iter)
         data = SubmissionSerializer(submission).data
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -3,10 +3,11 @@ Public interface for the submissions app.
 
 """
 import copy
+import itertools
 import logging
+import operator
 import json
 
-from collections import namedtuple
 from django.conf import settings
 from django.core.cache import cache
 from django.db import IntegrityError, DatabaseError
@@ -88,15 +89,6 @@ class SubmissionRequestError(SubmissionError):
             if field_errors is not None
             else {}
         )
-
-
-# For API stability and low memory usage, large result sets are returned using namedtuples.
-# They take up only as much memory as tuples, but fields are accessible by name,
-# so we can change the fields in the future if needed
-LatestSubmission = namedtuple(
-    "LatestSubmission",
-    ("student_id", "attempt_number", "submitted_at", "created_at", "answer"),
-)
 
 
 def create_submission(student_item_dict, answer, submitted_at=None, attempt_number=None):
@@ -384,37 +376,44 @@ def get_submissions(student_item_dict, limit=None):
 
 
 def get_all_submissions(course_id, item_id, item_type, read_replica=True):
-    """
-    For the given item, get the most recent submission for every student who has submitted.
+    """For the given item, get the most recent submission for every student who has submitted.
 
     This may return a very large result set! It is implemented as a generator for efficiency.
-    It yields namedtuples with the following fields:
-        student_id
-        attempt_number
-        submitted_at
-        created_at
-        answer
 
-    Cannot fail unless there's a database error, but may return an empty list.
+    Args:
+        course_id, item_id, item_type (string): The values of the respective student_item fields
+            to filter the submissions by.
+        read_replica (bool): If true, attempt to use the read replica database.
+            If no read replica is available, use the default database.
+
+    Yields:
+        Dicts representing the submissions with the following fields:
+            student_item
+            student_id
+            attempt_number
+            submitted_at
+            created_at
+            answer
+
+    Raises:
+        Cannot fail unless there's a database error, but may return an empty iterable.
     """
-    query = Submission.objects.filter(
+    submission_qs = Submission.objects
+    if read_replica:
+        submission_qs = _use_read_replica(submission_qs)
+    query = submission_qs.select_related('student_item').filter(
         student_item__course_id=course_id,
         student_item__item_id=item_id,
         student_item__item_type=item_type,
-    ).values_list(
-        'student_item__student_id', 'attempt_number', 'submitted_at', 'created_at', 'raw_answer'
-    ).order_by('student_item__student_id', '-created_at')
+    ).order_by('student_item__student_id', '-created_at').iterator()
 
-    last_student_id = None
-    for row in query.all():
-        # We cannot use SELECT DISTINCT ON because it's PostgreSQL only, so unfortunately
-        # our results may contain every entry of each student, not just the most recent.
-        if row[0] == last_student_id:
-            continue  # Skip this row; it's an old submission for a student we've already included
-        last_student_id = row[0]
-        # Parse 'raw_answer' as JSON:
-        answer = json.loads(row[4])
-        yield LatestSubmission(row[0], row[1], row[2], row[3], answer)
+    # We cannot use SELECT DISTINCT ON because it's PostgreSQL only, so unfortunately
+    # our results will contain every entry of each student, not just the most recent.
+    for unused_student_id, row_iter in itertools.groupby(query, operator.attrgetter('student_item.student_id')):
+        submission = next(row_iter)
+        data = SubmissionSerializer(submission).data
+        data['student_id'] = submission.student_item.student_id
+        yield data
 
 
 def get_top_submissions(course_id, item_id, item_type, number_of_top_scores, use_cache=True, read_replica=True):
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -6,6 +6,7 @@ import copy
 import logging
 import json
 
+from collections import namedtuple
 from django.conf import settings
 from django.core.cache import cache
 from django.db import IntegrityError, DatabaseError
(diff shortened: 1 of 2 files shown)


edx-webhook:
Thanks for the pull request, @bradenmacdonald! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-submissions&number=14

submissions/api.py
```diff
@@ -373,6 +383,40 @@ def get_submissions(student_item_dict, limit=None):
     return SubmissionSerializer(submission_models, many=True).data
 
 
+def get_all_submissions(course_id, item_id, item_type, read_replica=True):
```
ormsbee: Would we ever have a read replica and choose not to use it?

submissions/api.py
```diff
@@ -373,6 +383,40 @@ def get_submissions(student_item_dict, limit=None):
...
+
+    Cannot fail unless there's a database error, but may return an empty list.
+    """
+    query = Submission.objects.filter(
+        student_item__course_id=course_id,
+        student_item__item_id=item_id,
+        student_item__item_type=item_type,
+    ).values_list(
+        'student_item__student_id', 'attempt_number', 'submitted_at', 'created_at', 'raw_answer'
+    ).order_by('student_item__student_id', '-created_at')
```
ormsbee: Please make sure:

1. We have a test that asserts that only one query is happening here (so we're not doing n+1 fetches for the join)
2. The query plan for this looks sane for a large data set. If it's going to have to do a file sort or something like that, it might be best to just do the tracking of "this was the most recent entry for this user" in memory. You can still make sure that you only show only the most recent by doing an order_by on the `-id` for the Submission model, and only showing the first one that pops up. This wouldn't give us a nice ordering by student ID, but I think that's a worthwhile tradeoff if it makes the query a lot faster.

submissions/api.py
bradenmacdonald: I don't know... but all of the methods here seem to have it as an option.

bradenmacdonald: Also, note to self: Need to add in the code to actually enable the read replica for this query.

submissions/api.py
```diff
@@ -373,6 +383,40 @@ def get_submissions(student_item_dict, limit=None):
...
+
+    last_student_id = None
+    for row in query.all():
+        # We cannot use SELECT DISTINCT ON because it's PostgreSQL only, so unfortunately
+        # our results may contain every entry of each student, not just the most recent.
+        if row[0] == last_student_id:
+            continue  # Skip this row; it's an old submission for a student we've already included
+        last_student_id = row[0]
+        # Parse 'raw_answer' as JSON:
+        answer = json.loads(row[4])
```
ormsbee: I don't think we want to impose the notion that all submissions are JSON blobs. Maybe that could be a method on the namedtuple? So something like:

```python
SUBMISSION_FIELDS = ["student_id", "attempt_number", "submitted_at", "created_at", "answer"]
class LatestSubmission(namedtuple('LatestSubmission', SUBMISSION_FIELDS)):
    __slots__ = ()
    def json(self):
        return json.loads(self.answer)

# then later in code...

seen_students = set()
submissions = (LatestSubmission(row) for row in query.all())
for sub in submissions:
    if sub.student_id not in seen_students:
        seen_students.add(sub.student_id)
        yield sub
```

submissions/api.py
```diff
@@ -89,6 +90,15 @@ def __init__(self, msg="", field_errors=None):
         )
 
 
+# For API stability and low memory usage, large result sets are returned using namedtuples.
+# They take up only as much memory as tuples, but fields are accessible by name,
+# so we can change the fields in the future if needed
+LatestSubmission = namedtuple(
```
ormsbee: LatestSubmission seems odd as a class name, given that it's just a submission. If we want to differentiate it from the model, I guess we could call it a SubmissionTuple or something.

submissions/api.py
bradenmacdonald: According to [a comment in `models.py`](https://github.com/edx/edx-submissions/blob/835b79040/submissions/models.py#L101), `raw_answer` is supposed to be JSON serialized.

I do like the idea of deferring the JSON parsing though.

ormsbee:
I have some concerns on the specifics, but overall I think this would be a very good addition. @wedaly: How does submissions data export work now for ORA2? I didn't see it in the code.

submissions/api.py
bradenmacdonald: I like `SubmissionTuple`, sure.

ormsbee:
Actually, thinking on this a little more, isn't there a valid case for downloading all submissions as well as the most recent ones? If we're talking about data-dumps in general?

bradenmacdonald:
@ormsbee  Good point. I can add that as a flag on the `get_all_submissions` method. (e.g. `latest_only=True` as another parameter)

wedaly:
@ormsbee There's a management command for the data dump that generates CSV files and uploads them to S3:

https://github.com/edx/edx-ora2/blob/master/openassessment/management/commands/upload_oa_data.py

https://github.com/edx/edx-ora2/blob/master/openassessment/data.py

brianhw:
As for data packages for researchers, Victor wrote up a set of SQL queries late last year, but the PR has not been merged:  https://github.com/edx/analytics-exporter/pull/51.

edx-webhook:
Thanks for the pull request, @bradenmacdonald! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?number=14&repo=edx%2Fedx-submissions

edx-webhook:
Thanks for the pull request, @bradenmacdonald! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-submissions&number=14

edx-webhook:
Thanks for the pull request, @bradenmacdonald! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-submissions&number=14

edx-webhook:
Thanks for the pull request, @bradenmacdonald! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-submissions&number=14

submissions/api.py
smarnach: The parameter is used in the tests – not sure whether there is any other use case.  (And I've changed the code to actually consider this parameter.)

submissions/api.py
smarnach: I've added a test to make sure there is only one query.

submissions/api.py
smarnach: All answers are already assumed to be JSON-serialisable.  The API returns dictionaries of submissions to the outside worls, never actual model instances.  These dictionaries are generated from model instances using the serialisers in `serializers.py`.  The serialiser for `Submission` instances [assumes that the `raw_answe` field is JSON](https://github.com/edx/edx-submissions/blob/master/submissions/serializers.py#L66), and handles the decoding and encoding.  I've refactored the code to use this framework, as do all other functions in the API.

submissions/api.py
smarnach: I've removed the named tuple and used dictionaries instead, as do all other functions in the API.  The only use of the named tuple was in the newly added `get_all_submissions()` function, which is a lazy generator anyway, so it is easy for clients to just extract the data they need while iterating over the result, thereby avoiding excessive memory consumption.

smarnach:
@ormsbee Could you please take another look at this PR?

submissions/api.py
ormsbee: ok.

submissions/api.py
```diff
@@ -373,6 +375,44 @@ def get_submissions(student_item_dict, limit=None):
...
+        Cannot fail unless there's a database error, but may return an empty iterable.
+    """
+    submission_qs = Submission.objects
+    if read_replica:
+        submission_qs = _use_read_replica(submission_qs)
+    query = submission_qs.select_related('student_item').filter(
+        student_item__course_id=course_id,
+        student_item__item_id=item_id,
+        student_item__item_type=item_type,
+    ).order_by('student_item__student_id', '-created_at').iterator()
```
ormsbee: Is the `order_by` here expensive (e.g. will it do a file-sort on large courses)? If so, can we drop it?

ormsbee: Oh, I see, we're using it to group results to only return the most recent. In that case, I would recommend that you inspect the `id` as a tie-breaker -- it is possible for double-entries to have the exact `created_at` because MySQL `datetimes` only have second-level granularity. It is possible that this operation as written would be non-deterministic.

ormsbee: (Well, I think MySQL has a more granular type, but the one Django uses is coarse -- it's caused us problems in the past.)

ormsbee:
Minor issue regarding timestamp granularity and tie-breaking, but I'm good with this PR after that's addressed. Thank you.

submissions/api.py
```diff
@@ -373,6 +375,44 @@ def get_submissions(student_item_dict, limit=None):
...
+
+    # We cannot use SELECT DISTINCT ON because it's PostgreSQL only, so unfortunately
+    # our results will contain every entry of each student, not just the most recent.
+    for unused_student_id, row_iter in itertools.groupby(query, operator.attrgetter('student_item.student_id')):
+        submission = next(row_iter)
+        yield SubmissionSerializer(submission).data
```
bradenmacdonald: This change makes the data export block not work for me. Because `SubmissionSerializer(submission).data` looks like this:

```python
{
    'uuid': u'ec26ccac-f8fd-11e4-a87f-080027880ca6',
    'student_item': 2L,
    'attempt_number': 5L,
    'submitted_at': datetime.datetime(2015, 5, 12, 23, 23, 39, tzinfo=<UTC>),
    'created_at': datetime.datetime(2015, 5, 12, 23, 23, 39, tzinfo=<UTC>),
    'answer': u'neither'
}
```

But the latest code in the data export block needs this result to contain `student_id` (it must be in the result, since querying the student_item to look it up each time would be too expensive):

```python
for submission in sub_api.get_all_submissions(course_key_str, block_id, block_type):
    student_id = submission['student_id']
```

smarnach: As discussed on IRC, this was me messing up git commands.  The correct version has been pushed now.

bradenmacdonald:
:+1: from me if you fix the issue Dave pointed out.

submissions/api.py
smarnach: Thanks.  I changed the ordering to

   order_by('student_item__student_id', '-submitted_at', '-id')

The default ordering on the model is by `'-submitted_at', '-id'`, so i used this over `created_at`.

PR #17, opened 2015-08-21
Upgrade django-extensions to 1.5.5

wedaly:


wedaly:
@rlucioni please review.  This is pre-work for the Django upgrade.

rlucioni:
:+1:

PR #20, opened 2015-08-27
Upgrade djangorestframework to v3.1

wedaly:
* Upgrade DRF to v.3.1
* Use JSONField on the model instead of custom serializer field for the submission answer.

I'm going to hold off on merging this until ORA2 and edx-platform are also upgraded.  Otherwise, it isn't going to be possible to test this end-to-end.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -46,6 +46,7 @@ INSTALLED_APPS = (
 
     # Third party
     'django_extensions',
+    'south',
 
     # Test
     'django_nose',
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='edx-submissions',
-    version='0.0.8',
+    version='0.1.0',
     author='edX',
     description='An API for creating submissions and scores.',
     url='http://github.com/edx/edx-submissions.git',
--- a/submissions/models.py
+++ b/submissions/models.py
@@ -9,7 +9,6 @@ need to then generate a matching migration for it using:
     ./manage.py schemamigration submissions --auto
 
 """
-import json
 import logging
 
 from django.db import models, DatabaseError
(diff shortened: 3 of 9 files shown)


submissions/tests/test_api.py
```diff
@@ -183,12 +185,14 @@ def test_create_non_json_answer(self):
...
         submission = api.create_submission(STUDENT_ITEM, ANSWER_ONE)
         sub_model = Submission.objects.get(uuid=submission['uuid'])
-        sub_model.raw_answer = ''
-        sub_model.save()
+
+        # This should never happen, if folks are using the public API.
+        # Create a submission with a raw answer that is NOT valid JSON
+        query = "UPDATE submissions_submission SET raw_answer = '}' WHERE id = %s"
+        connection.cursor().execute(query, [str(sub_model.id)])
+        transaction.commit_unless_managed()
```
muzaffaryousaf: @wedaly Why we are using `transaction.commit_unless_managed()` here.

wedaly: https://docs.djangoproject.com/en/1.4/topics/db/sql/#executing-custom-sql-directly

muzaffaryousaf: Actually this is removed in django 1.8. Can we use something else but i did't get why we are using this at all.

settings.py
```diff
@@ -46,6 +46,7 @@
 
     # Third party
     'django_extensions',
+    'south',
```
andy-armstrong: Doesn't South go away by Django 1.8?

setup.py
```diff
@@ -33,7 +33,7 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='edx-submissions',
-    version='0.0.8',
+    version='0.1.0',
```
andy-armstrong: Is there a plan to get this to version 1? It seems like it has proved itself in production now.

submissions/models.py
```diff
@@ -99,7 +99,12 @@ class Submission(models.Model):
     created_at = models.DateTimeField(editable=False, default=now, db_index=True)
 
     # The answer (JSON-serialized)
-    raw_answer = models.TextField(blank=True)
+    # NOTE: previously, this field was a TextField named `raw_answer`.
```
andy-armstrong: Just to be certain, you are saying that raw_answer was a text field that stored JSON values, and thus there are no migration concerns to switching it to a JSONField.

submissions/serializers.py
```diff
@@ -1,58 +1,36 @@
 """
 Serializers are created to ensure models do not have to be accessed outside the
-scope of the Tim APIs.
+scope of the submissions API.
```
andy-armstrong: :-)

settings.py
wedaly: Yes, but this PR is just upgrading DRF, not Django.  There is another branch for Django 1.8 in this repo that switches to Django migrations: https://github.com/edx/edx-submissions/pull/23

setup.py
wedaly: We're not very consistent with our version numbers across repos :)

andy-armstrong:
:+1: from me, but if possible it would make sense to address @muzaffaryousaf's comment to simplify the Django upgrade.

submissions/models.py
wedaly: Exactly.

wedaly:
I talked to @muzaffaryousaf about this offline -- basically, until we upgrade Django, I need to use Django 1.4's transaction management.  He's updated this code on his branch, so it isn't blocking the Django 1.8 upgrade work.

PR #23, opened 2015-09-09
Upgrade Django to 1.8.

muzaffaryousaf:


requirements.txt
```diff
@@ -1,8 +1,7 @@
 dogapi==1.2.1
-django>=1.4,<1.5
+django==1.8
```
muzaffaryousaf: Do we need 1.8 i can see other versions to upto 1.8.5 ?

symbolist: You can make it `>=1.8,<1.9` like the pattern used earlier.

settings.py
```diff
@@ -28,6 +28,10 @@
 SITE_ID = 1
 USE_TZ = True
 
+from django.utils.crypto import get_random_string
+chars = 'abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)'
+SECRET_KEY = get_random_string(50, chars)
```
muzaffaryousaf: This should be in config files also we need to figure out the proper way to generate this for edx.

submissions/tests/test_api.py
```diff
@@ -190,9 +190,9 @@ def test_load_non_json_answer(self):
 
         # This should never happen, if folks are using the public API.
         # Create a submission with a raw answer that is NOT valid JSON
-        query = "UPDATE submissions_submission SET raw_answer = '}' WHERE id = %s"
-        connection.cursor().execute(query, [str(sub_model.id)])
-        transaction.commit_unless_managed()
```
muzaffaryousaf: `transaction.commit_unless_managed()` is not supported in Django 1.8

test-requirements.txt
```diff
@@ -1,5 +1,5 @@
 ddt==0.8.0
-django-nose==1.2
+django-nose==1.4.1
```
muzaffaryousaf: Will remove this after rebase.

muzaffaryousaf:
@symbolist @wedaly @macdiesel @muhammad-ammar Can you guys give it a look. Thanks.

macdiesel:
Remind me of the plan for migrations?

muzaffaryousaf:
@macdiesel Here is the [link](https://docs.djangoproject.com/en/1.8/topics/migrations/#upgrading-from-south) that i followed for migrations. I'm not sure if that's what you were asking.

settings.py
wedaly: For edx-submissions, this settings file is used only in the test suite.  You're right that we'll need to add this to `*.auth.json` in edx-platform.

wedaly:
@muzaffaryousaf I noticed that you opened this PR into my branch to upgrade Django Rest Framework.  We should talk about when we're planning to merge the 1.8 upgrade back to master.

Is the plan to keep the 1.8 upgrade on a feature branch until edx-platform is ready?  (Since edx-submissions is run in-process with edx-platform?)

muzaffaryousaf:
@wedaly This PR is to get the review without DRF changes that you made. I will reopen this PR against master once we merged DRF into master.

Yes, this will stay in feature branch until we update edx-platform and related repos.

wedaly:
The changes look good to me!  :+1:

muzaffaryousaf:
@wedaly Anything wrong with the mentioned approach or any suggestions ? Thanks

settings.py
```diff
@@ -55,6 +58,13 @@
     'submissions'
 )
 
+# TODO: These are removed from global defaults. Not sure we need here or not but i have added this to remove warning.
```
symbolist: Not sure about this: https://docs.djangoproject.com/en/1.8/ref/settings/#middleware-classes

symbolist:
Besides the one comment about the version in the requirements file. :+1:

settings.py
muzaffaryousaf: These middleware were removed from defaults. I'm leaving them as it without the TODO otherwise it will show warnings.

settings.py
muzaffaryousaf: These settings are used for only tests, i'm leaving this as it.

muzaffaryousaf:
Closing in favor of https://github.com/edx/edx-submissions/pull/26

PR #25, opened 2015-10-28
Add the ability to annotate scores.

dianakhuang:
## [TNL-3462](https://openedx.atlassian.net/browse/TNL-3462)

Add a new `ScoreAnnotation` model for annotating 'special' scores that may exist for reasons outside the typical workflow (for example: staff overrides).

### Testing
- [x] Unit, integration, acceptance tests as appropriate
- [x] Database migrations are backwards-compatible

### Reviewers
- [ ] Code review: @andy-armstrong 
- [x] Code review: @efischer19 

FYI: @maxrothman - we're adding a new table here, @ormsbee - just in case you wanted to give it a quick once-over.

### Post-review
- [ ] Squash commits

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -10,13 +10,13 @@ import json
 
 from django.conf import settings
 from django.core.cache import cache
-from django.db import IntegrityError, DatabaseError
+from django.db import IntegrityError, DatabaseError, transaction
 from dogapi import dog_stats_api
 
 from submissions.serializers import (
     SubmissionSerializer, StudentItemSerializer, ScoreSerializer
 )
-from submissions.models import Submission, StudentItem, Score, ScoreSummary, score_set, score_reset
+from submissions.models import Submission, StudentItem, Score, ScoreSummary, ScoreAnnotation, score_set, score_reset
 
 logger = logging.getLogger("submissions.api")
 
@@ -698,7 +698,8 @@ def reset_score(student_id, course_id, item_id):
         logger.info(msg)
 
 
-def set_score(submission_uuid, points_earned, points_possible):
+def set_score(submission_uuid, points_earned, points_possible,
+              annotation_creator=None, annotation_type=None, annotation_reason=None):
     """Set a score for a particular submission.
 
     Sets the score for a particular submission. This score is calculated
@@ -708,6 +709,11 @@ def set_score(submission_uuid, points_earned, points_possible):
         submission_uuid (str): UUID for the submission (must exist).
         points_earned (int): The earned points for this submission.
         points_possible (int): The total points possible for this particular student item.
+
+        annotation_creator (str): An optional field for recording who gave this particular score
+        annotation_type (str): An optional field for recording what type of annotation should be created,
+                                e.g. "staff_override".
+        annotation_reason (str): An optional field for recording why this score was set to its value.
 
     Returns:
         None
@@ -761,9 +767,19 @@ def set_score(submission_uuid, points_earned, points_possible):
     # even though we cannot retrieve it.
     # In this case, we assume that someone else has already created
     # a score summary and ignore the error.
+    # TODO: once we're using Django 1.8, use transactions to ensure that these
+    # two models are saved at the same time.
     try:
         score_model = score.save()
         _log_score(score_model)
+        if annotation_creator is not None:
+            score_annotation = ScoreAnnotation(
+                score=score_model,
+                creator=annotation_creator,
+                annotation_type=annotation_type,
+                reason=annotation_reason
+            )
+            score_annotation.save()
         # Send a signal out to any listeners who are waiting for scoring events.
         score_set.send(
             sender=None,
--- a/submissions/models.py
+++ b/submissions/models.py
@@ -11,6 +11,7 @@ need to then generate a matching migration for it using:
 """
 import logging
 
+from south.modelsinspector import add_introspection_rules
 from django.db import models, DatabaseError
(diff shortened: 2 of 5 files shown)


submissions/api.py
```diff
@@ -709,6 +710,11 @@ def set_score(submission_uuid, points_earned, points_possible):
         points_earned (int): The earned points for this submission.
         points_possible (int): The total points possible for this particular student item.
 
+        score_creator (str): An optional field for recording who made this particular score
```
efischer19: Nit: "who made this particular score" reads a bit ambiguously to me - it could mean either:
  * The original learner, as in "Little Bobby made a score of 94 on his homework"
  * The grader, as in "Big Bobby made a score of 94; that is what will be assigned to Little Bobby's homework response"

Maybe "who gave this particular score"?

submissions/api.py
```diff
@@ -764,6 +770,14 @@ def set_score(submission_uuid, points_earned, points_possible):
     try:
         score_model = score.save()
         _log_score(score_model)
+        if score_creator is not None:
```
efischer19: What's the relationship between annotated and non-annotated scores? Is it possible for a score to exist, but not be annotated? Using `score_creator` in this conditional seems to imply that any score that has been created will be annotated.

If that's not the case, could we rename this variable to `annotation_creator`, or something else that more clearly illustrates the relationship between the annotator, the original score, and the annotated score?

efischer19:
Looks great, just a few questions about documentation and variable naming. Otherwise this looks like exactly what we need.

efischer19:
:+1:

andy-armstrong:
@dianakhuang Do you want to have a feature branch for this change, or are you thinking it is it safe to just merge it ahead of the full staff grading feature?

dianakhuang:
@andy-armstrong I think it's safe to merge it into edx-submissions and then cut a new release of edx-submissions when we're ready to use this feature.

submissions/api.py
```diff
@@ -764,6 +770,14 @@ def set_score(submission_uuid, points_earned, points_possible):
...
         score_model = score.save()
         _log_score(score_model)
+        if annotation_creator is not None:
+            score_annotation = ScoreAnnotation(
+                score=score_model,
+                creator=annotation_creator,
+                annotation_type=annotation_type,
+                reason=annotation_reason
+            )
+            score_annotation.save()
```
andy-armstrong: What should happen if there is an error saving the annotation? Is this whole operation transactional, or will we end up with a score without an annotation?

dianakhuang: We can end up with a score without an annotation if the annotation fails. Do you think it makes sense to make it a manual transaction where both have to succeed before commiting?

andy-armstrong: I think it would make sense to validate the score annotation before saving the score, so that if there is an issue (say one of the field values being too long) it is caught first.

submissions/tests/test_api.py
```diff
@@ -268,6 +271,28 @@ def test_set_score_signal(self, send_mock):
...
+        # for auditing purposes.
+        annotations = ScoreAnnotation.objects.all()
+        self.assertGreater(len(annotations), 0)
+        annotation = annotations[0]
+        self.assertEqual(annotation.score.points_earned, 11)
+        self.assertEqual(annotation.score.points_possible, 12)
+        self.assertEqual(annotation.annotation_type, annotation_type)
+        self.assertEqual(annotation.creator, creator_uuid)
+        self.assertEqual(annotation.reason, reason)
+
```
andy-armstrong: I'd add a test where the score annotation is invalid (say the annotation_type is too long) and verify the behavior.

submissions/models.py
```diff
@@ -274,3 +274,16 @@ def update_score_summary(sender, **kwargs):
...
+class ScoreAnnotation(models.Model):
+    """ Annotate individual scores with extra information if necessary. """
+
+    score = models.ForeignKey(Score)
+    # A string that will represent the 'type' of annotation,
+    # e.g. staff_override, etc.
+    annotation_type = models.CharField(max_length=255, blank=False, db_index=True)
+
+    # The anonymized User ID that the XBlock sees, not their real ID.
+    creator = models.CharField(max_length=255, blank=False, db_index=True)
```
andy-armstrong: It might be good to introduce an AnonymizedUserIDField so that we can share validation. I'm presuming we also store anonymized ids in the submissions themselves.

submissions/api.py
andy-armstrong: Yes, I do think it would make sense to keep it transactional if we can.

andy-armstrong:
@dianakhuang I'm done with my first pass and it looks great. I would like to see the operations be transactional and have tests for more of the error cases, but then I think this looks good to go.

dianakhuang:
@andy-armstrong Could you take another look? I didn't implement the transactions or write a test case where the annotation fails to save.

submissions/models.py
```diff
@@ -41,7 +55,7 @@ class StudentItem(models.Model):
 
     """
     # The anonymized Student ID that the XBlock sees, not their real ID.
-    student_id = models.CharField(max_length=255, blank=False, db_index=True)
+    student_id = AnonymizedUserIDField()
```
andy-armstrong: Am I remembering correctly that this doesn't require any migration, because the underlying storage is the same.

dianakhuang: Yes. To South, it's basically the same thing.

andy-armstrong:
:+1: Looks great, thanks

PR #32, opened 2016-01-28
Delete Student State ORA fix (Proof-of-concept)

efischer19:
Not at all the final version of this, but I have tested this locally,
and it deletes what we need to delete and seems to solve all the problems
described in TNL-3880.

Pairs with https://github.com/edx/edx-platform/pull/11371

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -643,6 +643,37 @@ def get_latest_score_for_submission(submission_uuid, read_replica=False):
     return ScoreSerializer(score).data
 
 
+def delete_student_item(student_id, course_id, item_id):
+    """
+    EFischer - dev testing this on my devstack, will document more fully after trying proof-of-concept
+    """
+    # Retrieve the student item
+    try:
+        student_item = StudentItem.objects.get(
+            student_id=student_id, course_id=course_id, item_id=item_id
+        )
+        student_item.delete()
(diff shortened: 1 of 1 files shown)


submissions/api.py
```diff
@@ -643,6 +643,37 @@ def get_latest_score_for_submission(submission_uuid, read_replica=False):
...
+def delete_student_item(student_id, course_id, item_id):
+    """
+    EFischer - dev testing this on my devstack, will document more fully after trying proof-of-concept
+    """
+    # Retrieve the student item
+    try:
+        student_item = StudentItem.objects.get(
+            student_id=student_id, course_id=course_id, item_id=item_id
+        )
+        student_item.delete()
```
efischer19: Note to reviewers: the default behavior in django is to cascade on delete (https://docs.djangoproject.com/en/1.8/ref/models/fields/, ctrl-F `on_delete`). Because of this behavior, this single lookup-then-delete will obliterate any trace of this item/student/user triplet in the submissions database.

One thing to investigate - how will this affect data in the `assessment` and `workflow` databases? Those entries are almost always keyed on `submission['uuid']`, so I'd imagine they'll just be orphaned, and possibly still included in a data dump? I'll keep digging there.

efischer19:
Closing in favor of https://github.com/edx/edx-submissions/pull/33

PR #33, opened 2016-02-01
Soft-delete Submissions to reset student state

efischer19:
Previous methods of "deleting state" for a student had just issued
reset scores, so the (unscored) submission was still hanging around.
With this change, the submission can be "deleted" for all django-
relevant purposes, while still remaining in the database in case
it turns out to be interesting to analytics later.

Supersedes https://github.com/edx/edx-submissions/pull/32.
Sibling change to https://github.com/edx/edx-platform/pull/11371 and https://github.com/edx/edx-ora2/pull/862

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -217,7 +217,7 @@ def get_submission(submission_uuid, read_replica=False):
             msg="submission_uuid ({!r}) must be a string type".format(submission_uuid)
         )
 
-    cache_key = "submissions.submission.{}".format(submission_uuid)
+    cache_key = Submission.get_cache_key(submission_uuid)
     try:
         cached_submission_data = cache.get(cache_key)
     except Exception as ex:
@@ -627,8 +627,10 @@ def get_latest_score_for_submission(submission_uuid, read_replica=False):
 
     """
     try:
+        # Ensure that submission_uuid is valid before fetching score
+        submission_model = Submission.objects.get(uuid=submission_uuid)
         score_qs = Score.objects.filter(
-            submission__uuid=submission_uuid
+            submission__uuid=submission_model.uuid
         ).order_by("-id").select_related("submission")
 
         if read_replica:
@@ -637,13 +639,13 @@ def get_latest_score_for_submission(submission_uuid, read_replica=False):
         score = score_qs[0]
         if score.is_hidden():
             return None
-    except IndexError:
+    except (IndexError, Submission.DoesNotExist):
         return None
 
     return ScoreSerializer(score).data
 
 
-def reset_score(student_id, course_id, item_id):
+def reset_score(student_id, course_id, item_id, clear_state=False):
     """
     Reset scores for a specific student on a specific problem.
 
@@ -655,6 +657,7 @@ def reset_score(student_id, course_id, item_id):
         student_id (unicode): The ID of the student for whom to reset scores.
         course_id (unicode): The ID of the course containing the item to reset.
         item_id (unicode): The ID of the item for which to reset scores.
+        clear_state (bool): If True, will appear to delete any submissions associated with the specified StudentItem
 
     Returns:
         None
@@ -683,6 +686,16 @@ def reset_score(student_id, course_id, item_id):
             course_id=course_id,
             item_id=item_id,
         )
+
+        if clear_state:
+            for sub in student_item.submission_set.all():
+                # soft-delete the Submission
+                sub.status = Submission.DELETED
+                sub.save(update_fields=["status"])
+
+                # Also clear out cached values
+                cache_key = Submission.get_cache_key(sub.uuid)
+                cache.delete(cache_key)
 
     except DatabaseError:
         msg = (
--- a/submissions/migrations/0003_submission_status.py
+++ b/submissions/migrations/0003_submission_status.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('submissions', '0002_auto_20151119_0913'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='submission',
+            name='status',
+            field=models.CharField(default=b'A', max_length=1, choices=[(b'D', b'Deleted'), (b'A', b'Active')]),
+        ),
+    ]
--- a/submissions/models.py
+++ b/submissions/models.py
@@ -124,6 +124,27 @@ class Submission(models.Model):
     # name so it continues to use `raw_answer`.
     answer = JSONField(blank=True, db_column="raw_answer")
 
+    # Has this submission been soft-deleted? This allows instructors to reset student
+    # state on an item, while preserving the previous value for potential analytics use.
+    DELETED = 'D'
+    ACTIVE = 'A'
+    STATUS_CHOICES = (
+        (DELETED, 'Deleted'),
+        (ACTIVE, 'Active'),
+    )
+    status = models.CharField(max_length=1, choices=STATUS_CHOICES, default=ACTIVE)
+
+    # Override the default Manager with our custom one to filter out soft-deleted items
+    class SoftDeletedManager(models.Manager):
+        def get_queryset(self):
+            return super(Submission.SoftDeletedManager, self).get_queryset().exclude(status=Submission.DELETED)
+
+    objects = SoftDeletedManager()
+
+    @staticmethod
+    def get_cache_key(sub_uuid):
+        return "submissions.submission.{}".format(sub_uuid)
+
     def __repr__(self):
         return repr(dict(
             uuid=self.uuid,
--- a/submissions/tests/test_api.py
+++ b/submissions/tests/test_api.py
@@ -573,6 +573,27 @@ class TestSubmissionsApi(TestCase):
             )
         self.assertEqual(cached_scores, scores)
 
+    def test_clear_state(self):
+        # Create a submission, give it a score, and verify that score exists
+        submission = api.create_submission(STUDENT_ITEM, ANSWER_ONE)
+        api.set_score(submission["uuid"], 11, 12)
+        score = api.get_score(STUDENT_ITEM)
+        self._assert_score(score, 11, 12)
+        self.assertEqual(score['submission_uuid'], submission['uuid'])
+
+        # Reset the score with clear_state=True
+        # This should set the submission's score to None, and make it unavailable to get_submissions
+        api.reset_score(
+            STUDENT_ITEM["student_id"],
+            STUDENT_ITEM["course_id"],
+            STUDENT_ITEM["item_id"],
+            clear_state=True,
+        )
+        score = api.get_score(STUDENT_ITEM)
+        self.assertIsNone(score)
+        subs = api.get_submissions(STUDENT_ITEM)
+        self.assertEqual(subs, [])
+
     @raises(api.SubmissionRequestError)
     def test_error_on_get_top_submissions_too_few(self):
         student_item = copy.deepcopy(STUDENT_ITEM)


submissions/models.py
```diff
@@ -33,6 +33,11 @@
     providing_args=['anonymous_user_id', 'course_id', 'item_id']
 )
 
+# In order to allow for effective soft-deletes, we want to override the default Manager for several object types
+class SoftDeletedManager(models.Manager):
+    def get_queryset(self):
+        #By default, filter out any soft-deleted records for classes using this manager
+        return super(SoftDeletedManager, self).get_queryset().filter(deleted=False)
```
efischer19: @mulby, I'm curious if this approach to soft-deleting will meet everyone's needs. By overriding the `objects` manager for the relevant classes, we'll essentially be hiding the deleted records from any query internal to the djangoapp.

When the data is gathered for analytics, is it accessed using django models (e.g. `Submissions.objects.all()`), or taken straight from the database using other methods? Is the same true for instuctor-initiated data downloads?

mulby: When we export to researchers we just dump the data directly out of the database using a MySQL CLI so it won't use the model.

I'm not sure about the downloadable reports. @ormsbee might know?

efischer19: Thanks @mulby, that's very helpful.

I found the PR where @cahrens is working on the data downloads here(https://github.com/edx/edx-platform/pull/11167), and it's also using SQL instead of models, so I think we'll be covered on all fronts.

submissions/api.py
```diff
@@ -637,12 +639,69 @@ def get_latest_score_for_submission(submission_uuid, read_replica=False):
...
+            cache_key = "submissions.submission.{}".format(sub.uuid)
+            cache.delete(cache_key)
+            sub.deleted = True
+            sub.save()
+
+        # ScoreSummary and StudentItem have uniqueness constraints that make soft-deleting problematic
+        # Luckily, creating a reset score will handle our needs adequately
+        Score.create_reset_score(student_item)
+
+        # TODO: the top scores cache is automatically invalidated after 5 minutes, should we rely on that instead of doing it manually here?
```
efischer19: Open question for reviewers here, the manual cache clearing is a bit unwieldy since `number` is included in the keys, is it worth doing anyways or should we let the cache expire?

submissions/models.py
```diff
@@ -33,6 +33,11 @@
     providing_args=['anonymous_user_id', 'course_id', 'item_id']
 )
 
+# In order to allow for effective soft-deletes, we want to override the default Manager for several object types
+class SoftDeletedManager(models.Manager):
+    def get_queryset(self):
+        #By default, filter out any soft-deleted records for classes using this manager
+        return super(SoftDeletedManager, self).get_queryset().exclude(deleted=True)
```
efischer19: @maxrothman, could we chat sometime about the performance implication of this approach? We're effectively adding a `WHERE NOT ('submissions_submission'.'deleted' = True)` clause to every query django makes to the `submissions` table. We're expecting a very large percentage of items to have `deleted` be False, the only ones that will be True are those the the instructor has individually chosen for use with the `Delete Student State` tool.

Does this set your devops-spidey senses tingling? Is there a way we could do this more intelligently?

Ticket for reference: https://openedx.atlassian.net/browse/TNL-3880

ormsbee:
Sorry for coming late and asking a stupid question, but why is it necessary to delete their submission? It isn't enough to delete the `submission_id` that is in their student module state for the problem?

efischer19:
That's actually a really good question @ormsbee - the impetus for this work is https://openedx.atlassian.net/browse/TNL-3880.

The reason this is a bug that's only now appearing is staff tools - this block allows us to look up a submission for a student given a username, without checking the student's module state: https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/staff_area_mixin.py#L338

Looking into it more just now, it seems that the reported buggy behavior only appears when using staff tools to "find" a "deleted" submission, this puts everything into a bad state (for the current staff user). However, refreshing the page gets everything back to where it needs to be.

So I suppose we could do a simple fix, and have that staff access codepath also check to ensure the submission is valid. But then again, we may end up in this position again if there's a different place that tries to get a "deleted" submission, since it's only "deleted" insofar as the student module drops it's record.

submissions/api.py
```diff
@@ -637,12 +639,69 @@ def get_latest_score_for_submission(submission_uuid, read_replica=False):
...
+        )
+
+        # There will *usually* only be one submission, but that's not guaranteed so "delete" all items
+        submissions = student_item.submission_set.all()
+        ret_vals = [sub.uuid for sub in submissions]
+        for sub in submissions:
+            cache_key = "submissions.submission.{}".format(sub.uuid)
+            cache.delete(cache_key)
+            sub.deleted = True
+            sub.save()
```
efischer19: This cache clearing needs to happen no matter which of the 3 unlinking options we end up choosing

efischer19:
This is the SQL that will be run with the new status column migration:
```
edxapp@precise64:~/edx-platform$ ./manage.py lms --settings=devstack sqlmigrate submissions 0003
BEGIN;
ALTER TABLE `submissions_submission` ADD COLUMN `status` varchar(1) DEFAULT 'A' NOT NULL;
ALTER TABLE `submissions_submission` ALTER COLUMN `status` DROP DEFAULT;

COMMIT;
```

submissions/migrations/0003_submission_status.py
```diff
@@ -0,0 +1,19 @@
...
+        ('submissions', '0002_auto_20151119_0913'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='submission',
+            name='status',
+            field=models.CharField(default=b'A', max_length=1, choices=[(b'D', b'Deleted'), (b'A', b'Active')]),
+        ),
+    ]
```
efischer19: @maxrothman (and FYI @scottrish), here's the timing data I got from my sandbox. 90 seconds to apply the migration to a table with 700,000 rows.

```
mysql> show create procedure stuff_submissions_tables;
+--------------------------+--------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------+----------------------+--------------------+
| Procedure                | sql_mode                                   | Create Procedure                                                                                                                                                                                                                                                                                                                                                                                       | character_set_client | collation_connection | Database Collation |
+--------------------------+--------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------+----------------------+--------------------+
| stuff_submissions_tables | STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION | CREATE DEFINER=`root`@`localhost` PROCEDURE `stuff_submissions_tables`(in numRows int)
begin
declare i int;
set i = 1;
start transaction;
while i < numRows do
insert into submissions_studentitem values(i, uuid(), uuid(), uuid(), "openassessment");
insert into submissions_submission values (i, uuid(), 1, now(), now(), concat(uuid(), uuid(), uuid()), i);
set i = i+1;
end while;
commit;
end | utf8                 | utf8_general_ci      | utf8_general_ci    |
+--------------------------+--------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------+----------------------+--------------------+
1 row in set (0.00 sec)

mysql> call stuff_submissions_tables(700000);
Query OK, 0 rows affected (1 min 41.62 sec)

mysql> select count(*) from submissions_submission;
+----------+
| count(*) |
+----------+
|   699999 |
+----------+
1 row in set (0.14 sec)

mysql> exit
Bye
[efischer19-sandbox] efischer19@efischer19 i-6fd3e6ea:~$ time sudo -u www-data /edx/bin/python.edxapp /edx/bin/manage.edxapp lms --settings aws migrate submissions 0003

Operations to perform:
  Target specific migration: 0003_submission_status, from submissions
Running migrations:
  Rendering model states... DONE
  Applying submissions.0003_submission_status... OK

real	1m29.847s
user	0m3.460s
sys	0m0.412s
```

efischer19: We're going to have to make a call here on feasibility (@e0d devops and @scottrish product).

Using the search linked below, we're looking at 500-1000 requests per hour, or on the order of tens per minute. @maxrothman and I have estimated that, given that request rate and the above timing data for a database not under load, we're looking at an upper bound of several hundred request failures during the migration. Is this an acceptable outage? It will only apply to users who are hitting the submissions database, which is only used by ORA and has a cache that will remain unaffected.

http://splunk.edx.org/en-US/app/search/search?earliest=-7d%40h&latest=now&q=search%20index%3D%22prod-edx%22%20sourcetype%3Dedx%20%22submissions.api%22%20NOT%20%22(cached)%22%20%7C%20timechart%20span%3D1h%20count%20by%20log_level&display.page.search.tab=visualizations&display.general.type=visualizations&sid=1455135927.4182068

efischer19:
@maxrothman @e0d @scottrish Here's the data I got from running the new no-default migration. It's about a 50% reduction in time, so more like 2 minutes than 3 for production (probably). I've tested, and this does still require a table lock,  causing errors for users making submissions during the migration. If we're not comfortable with this outage, we'll need to explore other options.
```
mysql> select count(*) from submissions_submission;
+----------+
| count(*) |
+----------+
|   700000 |
+----------+
1 row in set (0.12 sec)

mysql> select * from submissions_submission limit 10;
+----+--------------------------------------+----------------+----------------------------+----------------------------+--------------------------------------------------------------------------------------------------------------+-----------------+
| id | uuid                                 | attempt_number | submitted_at               | created_at                 | raw_answer                                                                                                   | student_item_id |
+----+--------------------------------------+----------------+----------------------------+----------------------------+--------------------------------------------------------------------------------------------------------------+-----------------+
|  1 | a52dbedc-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a52dbef6-d027-11e5-b22d-0a4b0a301c3fa52dbef9-d027-11e5-b22d-0a4b0a301c3fa52dbf01-d027-11e5-b22d-0a4b0a301c3f |               1 |
|  2 | a52dc2c6-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a52dc2d0-d027-11e5-b22d-0a4b0a301c3fa52dc2d3-d027-11e5-b22d-0a4b0a301c3fa52dc2d6-d027-11e5-b22d-0a4b0a301c3f |               2 |
|  3 | a52dc48d-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a52dc495-d027-11e5-b22d-0a4b0a301c3fa52dc497-d027-11e5-b22d-0a4b0a301c3fa52dc499-d027-11e5-b22d-0a4b0a301c3f |               3 |
|  4 | a52dc621-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a52dc628-d027-11e5-b22d-0a4b0a301c3fa52dc62a-d027-11e5-b22d-0a4b0a301c3fa52dc62c-d027-11e5-b22d-0a4b0a301c3f |               4 |
|  5 | a5348fb8-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a5348fcf-d027-11e5-b22d-0a4b0a301c3fa5348fd1-d027-11e5-b22d-0a4b0a301c3fa5348fd6-d027-11e5-b22d-0a4b0a301c3f |               5 |
|  6 | a53491ae-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a53491b6-d027-11e5-b22d-0a4b0a301c3fa53491b8-d027-11e5-b22d-0a4b0a301c3fa53491bb-d027-11e5-b22d-0a4b0a301c3f |               6 |
|  7 | a5376e32-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a5376e3f-d027-11e5-b22d-0a4b0a301c3fa5376e41-d027-11e5-b22d-0a4b0a301c3fa5376e46-d027-11e5-b22d-0a4b0a301c3f |               7 |
|  8 | a5377060-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a5377068-d027-11e5-b22d-0a4b0a301c3fa537706a-d027-11e5-b22d-0a4b0a301c3fa537706d-d027-11e5-b22d-0a4b0a301c3f |               8 |
|  9 | a53771ed-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a53771f4-d027-11e5-b22d-0a4b0a301c3fa53771f6-d027-11e5-b22d-0a4b0a301c3fa53771f8-d027-11e5-b22d-0a4b0a301c3f |               9 |
| 10 | a5377364-d027-11e5-b22d-0a4b0a301c3f |              1 | 2016-02-10 18:53:59.000000 | 2016-02-10 18:53:59.000000 | a537736b-d027-11e5-b22d-0a4b0a301c3fa537736d-d027-11e5-b22d-0a4b0a301c3fa537736f-d027-11e5-b22d-0a4b0a301c3f |              10 |
+----+--------------------------------------+----------------+----------------------------+----------------------------+--------------------------------------------------------------------------------------------------------------+-----------------+
10 rows in set (0.00 sec)

mysql> exit
Bye
[efischer19-sandbox] efischer19@efischer19 i-6fd3e6ea:~$ sudo -u www-data /edx/bin/python.edxapp /edx/bin/manage.edxapp lms --settings aws sqlmigrate submissions 0003
BEGIN;
ALTER TABLE `submissions_submission` ADD COLUMN `status` varchar(1) NULL;

COMMIT;
[efischer19-sandbox] efischer19@efischer19 i-6fd3e6ea:~$ time sudo -u www-data /edx/bin/python.edxapp /edx/bin/manage.edxapp lms --settings aws migrate submissions 0003

Operations to perform:
  Target specific migration: 0003_submission_status, from submissions
Running migrations:
  Rendering model states... DONE
  Applying submissions.0003_submission_status... OK

real	1m7.989s
user	0m3.468s
sys	0m0.424s

**Comment by ormsbee** [comment]
Can we just add a new table, do a join, and only have values in that table
for non-default statuses?

**Comment by maxrothman** [comment]
@efischer19 could you post the SQL the new migration runs?

**Comment by efischer19** [comment]
Sorry @maxrothman, meant to include that above.
```
[efischer19-sandbox] efischer19@efischer19 i-6fd3e6ea:~$ sudo -u www-data /edx/bin/python.edxapp /edx/bin/manage.edxapp lms --settings aws sqlmigrate submissions 0003

BEGIN;
ALTER TABLE `submissions_submission` ADD COLUMN `status` varchar(1) NULL;

COMMIT;
```
@ormsbee, we discussed your idea in planning this morning, and have decided that the amount of data isn't large enough to justify an extension table. From a product point of view, the small outage is acceptable.

The plan is now:
[ ] confirm with @maxrothman the extent of the service outage (currently estimated 2-3 minutes, maybe ~200 requests)
[ ] confirm behavior during the outage (should be a small server error, nothing that corrupts user's state, and able to be retried successfully after the window closes)
[ ] put up a warning banner ahead of the migration announcing what is happening
[ ] do the migration
[ ] take the banner down

**Comment by efischer19** [comment]
@dianakhuang @robrap This PR is ready for review, can you give it a look when you get a chance?

FYI @cahrens - this goes with the other 2 you've already reviewed, feel free to skip it if you're busy with Fedx stuff though.

The problems with the global status message in https://openedx.atlassian.net/browse/TNL-4048 have been fixed via https://github.com/edx/edx-platform/pull/11525, which is going out in this week's release so that we can move forward with this release plan next week.

File: submissions/api.py [hunk 6]
```diff
@@ -684,6 +686,23 @@ def reset_score(student_id, course_id, item_id):
...
         )
 
+        if clear_state:
+            for sub in student_item.submission_set.all():
+                # soft-delete the Submission
+                sub.status = Submission.DELETED
+                sub.save(update_fields=["status"])
+
+                # Also clear out cached values
+                cache_key = "submissions.submission.{}".format(sub.uuid)
```
cahrens: I would prefer just letting the 5 minute timeout take care of this.

submissions/migrations/0003_submission_status.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('submissions', '0002_auto_20151119_0913'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='submission',
+            name='status',
+            field=models.CharField(default=None, max_length=1, null=True, choices=[(b'D', b'Deleted'), (b'A', b'Active')]),
```
cahrens: Should the default value be 'Active'?

submissions/models.py
```diff
@@ -124,6 +123,25 @@ class Submission(models.Model):
...
+    # Has this submission been soft-deleted? This allows instructors to reset student
+    # state on an item, while preserving the previous value for potential analytics use.
+    DELETED = 'D'
+    ACTIVE = 'A'  # Note that, for migration purposes, a status value of None will be handled as 'Active'.
+    # If you are adding more statuses, be *sure* to write a data migration first to handle this!
+    STATUS_CHOICES = (
+        (DELETED, 'Deleted'),
+        (ACTIVE, 'Active'),
+    )
+    status = models.CharField(max_length=1, choices=STATUS_CHOICES, null=True, default=None)
```
efischer19: @robrap This is the decision Christina was deferring to you. Using a null default here is arguably not the "right thing" to do (versus setting Active as the default), but it does provide a speedup in the outage-inducing migration (in tests, 90 seconds vs 60, when run with no load, see previous comments on this PR. Devops estimated about a 2x increase when running on the loaded prod server)

submissions/api.py
robrap: I'd have a function like `get_submission_cache_key` (or `make...` or `create...`), following whatever naming convention we've established for something like this.  This function should be used all places the cache key is needed.

submissions/api.py
```diff
@@ -684,6 +686,23 @@ def reset_score(student_id, course_id, item_id):
...
+                cache.delete(cache_key)
+                # TODO: the top scores cache is automatically invalidated after 5 minutes, should we rely on that instead of doing it manually here?
+                cache_keys = ["submissions.top_submissions.{course}.{item}.{type}.{number}".format(
```
robrap: Same here.  Another method for this to create the individual cache key.

submissions/migrations/0003_submission_status.py
robrap: As discussed, since we are warning about migration time anyway, we'll go for Active as default.

submissions/models.py
```diff
@@ -124,6 +123,25 @@ class Submission(models.Model):
     # name so it continues to use `raw_answer`.
     answer = JSONField(blank=True, db_column="raw_answer")
 
+    # Has this submission been soft-deleted? This allows instructors to reset student
+    # state on an item, while preserving the previous value for potential analytics use.
+    DELETED = 'D'
+    ACTIVE = 'A'  # Note that, for migration purposes, a status value of None will be handled as 'Active'.
```
robrap: Update comments as needed for migration default change.

submissions/migrations/0003_submission_status.py
efischer19: Works for me.

FYI @maxrothman and @scottrish, per reviewers, we feel that getting this done "correctly" and setting the default to be active is worth slowing down the migration a bit. So we're back to the migration that took 90 seconds on a sandbox, likely closer to ~3 minutes on production.

submissions/models.py
```diff
@@ -124,6 +123,25 @@ class Submission(models.Model):
...
+    # If you are adding more statuses, be *sure* to write a data migration first to handle this!
+    STATUS_CHOICES = (
+        (DELETED, 'Deleted'),
+        (ACTIVE, 'Active'),
+    )
+    status = models.CharField(max_length=1, choices=STATUS_CHOICES, null=True, default=None)
+
+    # Override the default Manager with our custom one to filter out soft-deleted items
+    class SoftDeletedManager(models.Manager):
```
robrap: Maybe rename to something like `ExcludeDeletedManager`?  Not sure about the name, but the current name sounds like it would provide the DELETED records.

robrap:
Looks good @efischer19.  All of my comments are minor.  Thanks.

submissions/api.py
efischer19: Just to ensure we're clear - the individual submission cache (keyed on `submissions.submission.{}`) does not have a timeout, and needs to be handled here. The top scores cache is what would timeout, and I agree that it probably doesn't need to be explicitly handled here.

efischer19:
@cahrens @robrap comments have been addressed

submissions/api.py
```diff
@@ -637,13 +639,13 @@ def get_latest_score_for_submission(submission_uuid, read_replica=False):
...
             return None
-    except IndexError:
+    except (IndexError, Submission.DoesNotExist):
         return None
 
     return ScoreSerializer(score).data
 
 
-def reset_score(student_id, course_id, item_id):
+def reset_score(student_id, course_id, item_id, clear_state=False):
```
dianakhuang: Could you document the differences in behavior `clear_state` causes? It's not immediately clear because of the subtleties of how the submissions tables work.

dianakhuang:
Other than that documentation update, this looks good. :+1:

PR #34, opened 2016-02-04
Clear state option

efischer19:
To fix the buggy behavior reported in TNL-3880, we need to ensure
that the submission we're dropping is unlinked from the student_item
used to find it. This prevents code (say, an ORA staff tool) from
constructing a student item with an id, a course, and a problem, and
using that to find a "cleared" submission.

[TNL-3880](https://openedx.atlassian.net/browse/PROJ-TNL-3880)
Sibling changes:
* https://github.com/edx/edx-platform/pull/11422
* https://github.com/edx/edx-ora2/pull/863

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -689,6 +689,7 @@ def reset_score(student_id, course_id, item_id, clear_state=False):
             # sever the link between this student item and any submissions it may current have
             for sub in student_item.submission_set.all():
                 sub.student_item = None
+                sub.save(update_fields=['student_item'])
 
     except DatabaseError:
         msg = (
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -643,7 +643,7 @@ def get_latest_score_for_submission(submission_uuid, read_replica=False):
     return ScoreSerializer(score).data
 
 
-def reset_score(student_id, course_id, item_id):
+def reset_score(student_id, course_id, item_id, clear_state=False):
     """
     Reset scores for a specific student on a specific problem.
 
@@ -655,6 +655,7 @@ def reset_score(student_id, course_id, item_id):
         student_id (unicode): The ID of the student for whom to reset scores.
         course_id (unicode): The ID of the course containing the item to reset.
         item_id (unicode): The ID of the item for which to reset scores.
+        clear_state (boolean): If True, unlink the Submission and StudentItem so the Submission cannot be accessed.
 
     Returns:
         None
@@ -683,6 +684,11 @@ def reset_score(student_id, course_id, item_id):
             course_id=course_id,
             item_id=item_id,
         )
+
+        if clear_state:
+            # sever the link between this student item and any submissions it may current have
+            for sub in student_item.submission_set.all():
+                sub.student_item = None
 
     except DatabaseError:
         msg = (
--- a/submissions/migrations/0003_submission_null_FK.py
+++ b/submissions/migrations/0003_submission_null_FK.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('submissions', '0002_auto_20151119_0913'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='submission',
+            name='student_item',
+            field=models.ForeignKey(to='submissions.StudentItem', null=True),
+        ),
+    ]
--- a/submissions/migrations/0003_submission_null_FK.py
+++ b/submissions/migrations/0003_submission_null_FK.py
@@ -14,6 +14,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='submission',
             name='student_item',
-            field=models.ForeignKey(to='submissions.StudentItem', null=True),
+            field=models.ForeignKey(to='submissions.StudentItem', null=True, default=1),
+            preserve_default=False
         ),
     ]
--- a/submissions/models.py
+++ b/submissions/models.py
@@ -103,7 +103,7 @@ class Submission(models.Model):
 
     uuid = UUIDField(version=1, db_index=True)
 
-    student_item = models.ForeignKey(StudentItem)
+    student_item = models.ForeignKey(StudentItem, null=True)
 
     # Which attempt is this? Consecutive Submissions do not necessarily have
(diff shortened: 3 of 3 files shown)


edx-webhook:
# PR Cover Letter

# Sandbox

# Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

# Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] UI strings/error msgs review
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

# DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

# Post-review
- [ ] Squash commits

efischer19:
@dianakhuang @cahrens This simpler fix is ready for an initial review (not final, still some open questions to answer).

What I misunderstood originally was that we aren't actually deleting anything, we're just orphaning submissions by having a student's module state forget the submission_uuid. However, staff can construct a perfectly valid lookup triplet of (student, course, problem), use that to query the submissions API, and get a (supposedly "orphaned") submission_uuid back.

This fix will now eliminate the key that made such lookups possible, which prevents staff from seeing the individual submission even if they do construct the lookup triplet again. This solves the problem of staff being able to see deleted items with the `Manage Individual Learners` tool.

What this does not address is the aggregate data in the `View Assignment Statistics` tool. This is looking up data (in an ORA database) keyed on (course, problem). Since ORA was never informed of the 'reset state' request, it has no way of knowing that these 'deleted' values should be removed. I have several ideas of how to fix this, but I also wonder if it's worthwhile. How do others feel? To fix it would require informing ORA of "cleared" submissions, either by the LMS at time of deletion, or by submissions at lookup time.

efischer19:
@maxrothman what's the devops impact of an AlterField migration? We're just removing a `NOT NULL` constraint from a column here.

efischer19:
To go along with the `View Assignment Statistics` issue mentioned above - ORA also stores references to the should-have-been-deleted submission_uuid in various workflow items, meaning that peers (and I assume staff, have yet to check) can be given previously "cleared" submissions for grading.

So it looks like we will need to inform ORA of the removal somehow, so it can clear out any items that refer to the cleared submission. I'll keep working on that.

submissions/migrations/0003_submission_null_FK.py
```diff
@@ -0,0 +1,19 @@
...
+        ('submissions', '0002_auto_20151119_0913'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='submission',
+            name='student_item',
+            field=models.ForeignKey(to='submissions.StudentItem', null=True),
+        ),
+    ]
```
efischer19: One potential issue I've discovered during dev testing - trying to reverse this migration after NULL values have been added results in an error, so I'll have to figure that out, at least.

`django.db.utils.OperationalError: (1138, 'Invalid use of NULL value')`.

efischer19:
I've updated this PR and it's siblings to inform openassessment of the clear operation, and it looks like it's working.

@dianakhuang and/or @cahrens, could we start going over this in a bit more detail when we get a chance? I've installed all of this code on https://efischer19.sandbox.edx.org/ if you want to test it out.

submissions/migrations/0003_submission_null_FK.py
```diff
@@ -0,0 +1,20 @@
...
+
+    operations = [
+        migrations.AlterField(
+            model_name='submission',
+            name='student_item',
+            field=models.ForeignKey(to='submissions.StudentItem', null=True, default=1),
+            preserve_default=False
+        ),
+    ]
```
efischer19: @maxrothman Here's some data from running this migration on my devstack with ~1400 rows of data, the read-replica indicates ~670,000 on production

https://gist.github.com/efischer19/b56b9d14768de28cf612

submissions/api.py
```diff
@@ -684,6 +685,12 @@ def reset_score(student_id, course_id, item_id):
             item_id=item_id,
         )
 
+        if clear_state:
+            # sever the link between this student item and any submissions it may current have
+            for sub in student_item.submission_set.all():
+                sub.student_item = None
+                sub.save(update_fields=['student_item'])
```
efischer19: @mulby We're a little curious about how this will impact analytics.

Essentially what happens on a "reset state" request is that we leave the `Submission` item in the database, and when the `StudentModule` is deleted (https://github.com/edx/edx-platform/blob/master/lms/djangoapps/instructor/enrollment.py#L246-L253), the only possible reference to the submission (via its uuid) is lost.

However, with the addition of staff tools, now a third party (the staff user) can construct a triplet of (user/course/problem location) and use *that* as a key to access a supposedly orphaned submission. The way we're attempting to address the issue in this PR is to null out the key for an orphaned submission, which will then be persisted in the database.

Will this be an issue for analytics? It will make it possible for `Submission` data to exist that cannot be easily linked to a student. Would adding events coverage make it any better?

efischer19: An alternative approach is https://github.com/edx/edx-submissions/pull/35, where we mangle the key instead of deleting it. The course and student information would remain intact, but the problem identifier is reversed.

submissions/migrations/0003_submission_null_FK.py
maxrothman: @efischer19 Do you feel those results are representative enough to indicate lack of an issue with ~100x rows?

submissions/api.py
mulby: I'm not sure how valuable it is to keep the data if it can't actually be associated with the original student anymore.

Are we confident we have good event log coverage of this?

efischer19: I'm not sure how thorough the event coverage is, but see my latest comment on https://github.com/edx/edx-submissions/pull/35. We may be able to keep the original student association mostly intact.

submissions/migrations/0003_submission_null_FK.py
efischer19: I think so. The two longest lines seem to be these, what else should we be considering when extrapolating?
```
DEBUG:django.db.backends:(0.114) ALTER TABLE `submissions_submission` MODIFY `student_item_id` integer NULL; args=[]
DEBUG:django.db.backends:(0.130) ALTER TABLE `submissions_submission` ADD CONSTRAINT `su_student_item_id_d3801ff833d05b1_fk_submissions_studentitem_id` FOREIGN KEY (`student_item_id`) REFERENCES `submissions_studentitem` (`id`); args=[]
```

There's also now a possibility of avoiding migrations entirely if the approach in https://github.com/edx/edx-submissions/pull/35 wins out as our best option.

PR #43, opened 2016-03-08
Use score summaries in data download

efischer19:
Let's see how this goes

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -447,27 +447,27 @@ def get_all_course_submission_information(course_id, item_type, read_replica=Tru
     if read_replica:
         submission_qs = _use_read_replica(submission_qs)
 
-    query = submission_qs.select_related('student_item').prefetch_related('score_set').filter(
+    query = submission_qs.select_related('student_item__scoresummary__latest__submission').filter(
         student_item__course_id=course_id,
         student_item__item_type=item_type,
     ).iterator()
 
     for submission in query:
         student_item = submission.student_item
-        if submission.score_set.count() > 0:
-            for score in submission.score_set.all():
-                yield (
-                    StudentItemSerializer(student_item).data,
-                    SubmissionSerializer(submission).data,
-                    ScoreSerializer(score).data
-                )
-        else:
-            # Make sure we return submission information even if there isn't a score associated with it.
-            yield (
-                StudentItemSerializer(student_item).data,
-                SubmissionSerializer(submission).data,
-                {}
-            )
+        serialized_score = {}
+        if hasattr(student_item, 'scoresummary'):
+            latest_score = student_item.scoresummary.latest
+
+            # We only include the score for a given submission if it is not a reset score and it "counts", that is,
+            # if it is the latest score on the score summary tracking the submission's student_item. This matches the
+            # behavior of the API's get_score method.
+            if (not latest_score.is_hidden()) and latest_score.submission.uuid == submission.uuid:
+                serialized_score = ScoreSerializer(latest_score).data
+        yield (
+            StudentItemSerializer(student_item).data,
+            SubmissionSerializer(submission).data,
+            serialized_score
+        )
 
 
 def get_top_submissions(course_id, item_id, item_type, number_of_top_scores, use_cache=True, read_replica=True):
--- a/submissions/tests/test_api.py
+++ b/submissions/tests/test_api.py
@@ -127,7 +127,8 @@ class TestSubmissionsApi(TestCase):
 
         self.assertDictEqual(SECOND_STUDENT_ITEM, submissions_and_scores[1][0])
         self._assert_submission(submissions_and_scores[1][1], submission3['answer'], student_item2.pk, 1)
-        self._assert_score(submissions_and_scores[1][2], 3, 4)
+        # submission4 also pertains to this student item and got its score later, so no score will be reported here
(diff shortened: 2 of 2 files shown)


edx-webhook:
## PR Cover Letter

## Sandbox

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] Docs review (required for UI strings/error messages)
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## Documentation considerations
- [ ] Are there user visible changes? If so, list who is affected, and include a pointer to the relevant code.
- [ ] Should this be mentioned in the release notes? If so, make sure that you included a description.

## DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

## Post-review
- [ ] Squash commits

submissions/api.py
```diff
@@ -447,27 +447,27 @@ def get_all_course_submission_information(course_id, item_type, read_replica=Tru
     if read_replica:
         submission_qs = _use_read_replica(submission_qs)
 
-    query = submission_qs.select_related('student_item').prefetch_related('score_set').filter(
+    query = submission_qs.select_related('student_item__scoresummary__latest__submission').filter(
```
efischer19: This `select_related` means that, for each `submission` in our query, we are fetching:
- its student_item, which can be tied to a
- scoresummary, which can be tied to a
- score (latest), which can be tied to a
- submission

The upshot is that, for each submission, we can ask the question `If we tried to get the score for this submission's student_item, would the score reported be associated with this submission?` If it is, the score is included in our data download. If it is not, no score is returned.

This is to handle a situation like this one:
- student makes a submission
- student gets a poor score on that submission
- student makes a new submission
- student gets a better score on that submission
The returned values in this case would be both submissions, but no score would be reported for the first one. Is this desirable? Or would it be better to return both scores even though one of them doesn't count? We have elected to go this route for now, since that seems to be how the api is set up for `get_score`.

Worth noting: ORA imposes a limit of 1 submission per student item (excluding delete state requests, but that data already excluded), so this is irrelevant to the context of ORA data downloads. It's more a question of how edx-submissions want to handle this theoretical case for other consumers in the future.

efischer19:
@dianakhuang Here's what I have for only including relevant scores in the data download.

@robrap Could you check over my comment above, and see if you agree with the route we're taking?

dianakhuang:
:+1:

submissions/api.py
robrap: As you stated, it really is a question of who needs the data download and what they want included:
1. Do they want all scores with additional data (like a latest flag) to help them filter the data in the download appropriately?  Or,
2. Does someone just want the latest (i.e. relevant) scores with no other historical scores?

One possibility is to go with the latest, since it is simpler, and add an option later if and when it is needed.  Can we find out if this aligns with the original requirement at a minimum?

@ormsbee @scottrish Thoughts on this?

Depending on which way we go, I'll have some additional minor requests on naming and commenting to make this more clear.

efischer19:
I don't want to close off this discussion, but we have the loadtest environment available today and would like to try this code out to get some ballpark figures.

I'm going to merge this into Diana's branch, but the question of "only latest/relevant scores" vs. "all scores with flag" is still open. If we do end up going with "all scores and a flag", the performance will probably be slightly worse, but not terribly so.

submissions/api.py
```diff
@@ -447,27 +447,27 @@ def get_all_course_submission_information(course_id, item_type, read_replica=Tru
...
-                SubmissionSerializer(submission).data,
-                {}
-            )
+        serialized_score = {}
+        if hasattr(student_item, 'scoresummary'):
+            latest_score = student_item.scoresummary.latest
+
+            # We only include the score for a given submission if it is not a reset score and it "counts", that is,
+            # if it is the latest score on the score summary tracking the submission's student_item. This matches the
+            # behavior of the API's get_score method.
```
robrap: @efischer19 @dianakhuang  Does the following comment sound more clear to you?
```
# Only include the score if it is not a reset score (is_hidden), and if the current submission matches 
# the latest score's submission.  This matches the behavior of the API's get_score method.
```

submissions/api.py
robrap: If we stick with latest score, the comment for this method should mention "latest score" each time it currently reads "score".  The function comment should also mention that an empty score will be sent for submissions with no score.

submissions/api.py
efischer19: It's better, but I'm still looking at `if the current submission matches the latest score's submission`. Is there some way we can mention the student_item in there as well? Something like `if the current submission corresponds to the latest score for this student_item`?

robrap: I'd like something that makes all the lines of code more clear, including the last test comparing the submission.uuid's.

Maybe: 
```
...if the current submission is the same as the student_item's latest score's submission.
```

efischer19: :+1: I like that one, I'll implement it when I update this code later. Thanks!

submissions/api.py
ormsbee: My guess is that most people will only want the latest relevant score with no historical data for something like a download, so that should be the default behavior. But I think there are valid reasons for wanting every submission and its corresponding score or scores in the future.

PR #49, opened 2016-11-07
Unpin dependencies in setup.py

cpennington:


jmbowman:
It looks like this has already effectively been done, but in a slightly different manner (the requirements file constraints were loosened); close the PR?

PR #50, opened 2016-12-02
Serialize annotation information with scores

efischer19:
This is to fix a bug where ORA cannot tell accurately what scores are from staff.

If this gets merged, I'll be sure to handle releasing as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='edx-submissions',
-    version='1.1.2',
+    version='1.1.3',
     author='edX',
     description='An API for creating submissions and scores.',
     url='http://github.com/edx/edx-submissions.git',
--- a/submissions/serializers.py
+++ b/submissions/serializers.py
@@ -6,7 +6,7 @@ import json
 
 from rest_framework import serializers
 from rest_framework.fields import Field, DateTimeField, IntegerField
-from submissions.models import StudentItem, Submission, Score
+from submissions.models import StudentItem, Submission, Score, ScoreAnnotation
 
 
 class RawField(Field):
@@ -85,10 +85,32 @@ class SubmissionSerializer(serializers.ModelSerializer):
         )
 
 
+class ScoreAnnotationSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = ScoreAnnotation
+        fields = (
+            'creator',
+            'reason',
+            'annotation_type',
+        )
+
+
 class ScoreSerializer(serializers.ModelSerializer):
 
     # Ensure that the created_at datetime is not converted to a string.
     created_at = DateTimeField(format=None, required=False)
+
+    annotations = serializers.SerializerMethodField()
+    def get_annotations(self, obj):
+        """
+        Inspect ScoreAnnotations to attach all relevant annotations.
+        """
+        annotations = ScoreAnnotation.objects.filter(score_id=obj.id)
+        return [
+            ScoreAnnotationSerializer(instance=annotation).data
+            for annotation in annotations
+        ]
 
     class Meta:
         model = Score
@@ -101,4 +123,5 @@ class ScoreSerializer(serializers.ModelSerializer):
 
             # Computed
             'submission_uuid',
+            'annotations',
         )
--- a/submissions/tests/test_serializers.py
+++ b/submissions/tests/test_serializers.py
@@ -1,32 +1,72 @@
 """
 Tests for submissions serializers.
 """
+import ddt
 from django.test import TestCase
-from submissions.models import Score, StudentItem
+from submissions.models import Score, ScoreAnnotation, StudentItem, Submission
 from submissions.serializers import ScoreSerializer
 
 
+@ddt.ddt
 class ScoreSerializerTest(TestCase):
     """
     Tests for the score serializer.
     """
 
-    def test_score_with_null_submission(self):
-        item = StudentItem.objects.create(
+    def setUp(self):
+        super(ScoreSerializerTest, self).setUp()
+        self.item = StudentItem.objects.create(
             student_id="score_test_student",
             course_id="score_test_course",
             item_id="i4x://mycourse/special_presentation"
         )
+        self.submission = Submission.objects.create(student_item=self.item, attempt_number=1)
 
-        # Create a score with a null submission
-        score = Score.objects.create(
-            student_item=item,
-            submission=None,
+        self.score = Score.objects.create(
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -33,7 +33,7 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='edx-submissions',
-    version='1.1.2',
+    version='1.1.3',
```
efischer19: Point of note: [searching for edx-submissions across all edx repos](https://github.com/search?utf8=%E2%9C%93&q=user%3Aedx+edx-submissions&type=Code&ref=searchresults) show that ora2 and edx-platform are the only locations that will need to have their pinned version bumped.

efischer19:
This PR is ready for review; it adds data about score annotation types to the default `ScoreSerializer`, which surfaces that information in calls like `get_latest_score_for_submission`. This is the first step in resolving the bug noted on [TNL-6092](https://openedx.atlassian.net/browse/TNL-6092), the second step will be to have ORA inspect that information and use it when determining whether a given score was staff-given or not (see https://github.com/edx/edx-ora2/pull/951/files).

Once this gets merged, I'll create a new version tag and update both edx-ora2 and edx-platform as needed to get the new version.

Could I get one of @staubina or @cahrens, along with one of @nasthagiri, @jcdyer or @sanfordstudent to be my reviewers here?

submissions/tests/test_serializers.py
```diff
@@ -1,32 +1,59 @@
...
-            points_earned=2,
-            points_possible=6
+            points_earned=3,
+            points_possible=8,
         )
-        score_dict = ScoreSerializer(score).data
+        null_sub_score_dict = ScoreSerializer(null_sub_score).data
+        self.assertIs(null_sub_score_dict['submission_uuid'], None)
+        self.assertEqual(null_sub_score_dict['points_earned'], 3)
+        self.assertEqual(null_sub_score_dict['points_possible'], 8)
```
nasthagiri: Why were the changes in this pre-existing test needed?

submissions/tests/test_serializers.py
```diff
@@ -1,32 +1,59 @@
...
 
-        self.assertIs(score_dict['submission_uuid'], None)
-        self.assertEqual(score_dict['points_earned'], 2)
-        self.assertEqual(score_dict['points_possible'], 6)
+    @ddt.data(['test_annotation_1', 'test_annotation_2'], [])
```
nasthagiri: Can an annotation_type ever be `None`?

submissions/serializers.py
```diff
@@ -101,4 +109,5 @@ class Meta:
 
             # Computed
             'submission_uuid',
+            'annotation_types',
```
nasthagiri: How big do we expect this to get?  It won't ever need to be paginated, right?

efischer19: I don't believe so. It's currently only used for the `staff_defined` use case.

submissions/tests/test_serializers.py
efischer19: Because there was only 1 test method in this file before, I took the liberty of modifying it to use a common `setUp` method. Then I wanted the specialized score being tested here to be defined differently than the generalized `self.score`, which is why it's 3/8 instead of 2/6.

submissions/tests/test_serializers.py
efischer19: No. The way I've defined `get_annotation_types` above is to generate a list comprehension using database results, if there are no results then an empty list is returned.

submissions/serializers.py
nasthagiri: I see.  Thanks.

submissions/serializers.py
```diff
@@ -90,6 +90,14 @@ class ScoreSerializer(serializers.ModelSerializer):
     # Ensure that the created_at datetime is not converted to a string.
     created_at = DateTimeField(format=None, required=False)
 
+    annotation_types = serializers.SerializerMethodField()
+    def get_annotation_types(self, obj):
+        """
+        Inspect ScoreAnnotations to attach the annotation_type of any found annotations.
+        """
+        annotations = ScoreAnnotation.objects.filter(score_id=obj.id)
+        return [annotation.annotation_type for annotation in annotations]
```
cahrens: I question making this a list. If there are actually multiple annotations, knowing which one is related to a specific "annotation_type" seems like it would be very useful. I would think that a dict would make more sense, with the key being specific to the annotation (I don't know what annotation looks like off-hand).

efischer19: A `ScoreAnnotation` consists of an `annotation_type`, a `reason`, and a `creator`, all linked to a specific `Score` via a FK relationship ([source](https://github.com/edx/edx-submissions/blob/master/submissions/models.py#L323-L335)). I didn't think the reason or creator needed to be exposed at this level, so just went with the `annotation_type`.

Do you think we'd be better off exposing all that information in a dict like this?
```
'annotations':
  [{
    'type': 'staff_defined',
    'reason': 'B/c I said so',
    'creator': 12345,
  }]
```

cahrens: Yes, think that would be better. Returning just the type seems overly specific for this one particular use case.

efischer19:
@cahrens I've updated the code per your comments if you want to take another look.

submissions/serializers.py
```diff
@@ -90,6 +90,21 @@ class ScoreSerializer(serializers.ModelSerializer):
     # Ensure that the created_at datetime is not converted to a string.
     created_at = DateTimeField(format=None, required=False)
 
+    annotations = serializers.SerializerMethodField()
+    def get_annotations(self, obj):
+        """
+        Inspect ScoreAnnotations to attach the annotation_type of any found annotations.
```
cahrens: Update the doc string.

efischer19: gah!

submissions/serializers.py
```diff
@@ -90,6 +90,21 @@ class ScoreSerializer(serializers.ModelSerializer):
...
+        """
+        annotations = ScoreAnnotation.objects.filter(score_id=obj.id)
+        return [
```
cahrens: Isn't there a default serializer for dicts? I would expect it to serialize whatever fields are present (so this code does not need to change in the number of fields in the dict changes in the future).

efischer19: Adding the companion `ScoreAnnotationSerializer` was the closest I could get here, is that similar to what you were thinking?

efischer19:
Squashed commits after discussing @cahrens' concerns offline

cahrens:
👍

PR #52, opened 2017-02-15
Unpin and inherit django version

efischer19:
TNL-6131


efischer19:
@robrap Is there anything else required to handle the issue flagged in https://openedx.atlassian.net/browse/TNL-6131? If not, I'll go about the process of getting this merged and released this afternoon.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 dogapi==1.2.1
-django>=1.8.14,<1.9
+django<1.9
```
jcdyer: Should keep a `>1.8` pin.  It would also be nice not to bound the top unnecessarily, but I guess we don't have tox in place to make sure we test a 1.8 version.

efischer19:
THAT'S what's getting Gemnasium all worked up! Thanks @robrap and @yro!

efischer19:
It looks like 1.9 and 1.10 currently do not pass tests, but the infrastructure is now in place when we do want to run it. [Details at this test link.](https://travis-ci.org/edx/edx-submissions/builds/204321658)

I've updated `setup.py` to require a version `<1.9a0` to make Gemnasium happy, per the original ticket.

@jcdyer and @robrap can you re-review?

setup.py
```diff
@@ -46,6 +46,6 @@ def load_requirements(*requirements_paths):
         'Programming Language :: Python',
     ],
     packages=['submissions', 'submissions.migrations'],
-    install_requires=load_requirements('requirements.txt'),
+    install_requires=load_requirements('requirements.txt').append('Django<1.9a0'),
```
efischer19: I'm working to make sure this syntax is recognized by gemnasium, will update when I get more details.

.travis.yml
```diff
@@ -1,6 +1,11 @@
 language: python
 python:
   - "2.7"
+env:
+  - TOXENV=django18
+  # 1.9 and 1.10 known to fail currently
+  #- TOXENV=django19
```
efischer19: Think it's worth doing this for 1.10 as well then? They both fail in different ways.

efischer19:
I just manually updated and confirmed Gemnasium is happy with the state of code on master; I'll keep the story open until a new release is made and merged into edx-platform.

PR #54, opened 2017-03-14
Allow reset signal to be ignored

efischer19:
# [TNL-6679](https://openedx.atlassian.net/browse/TNL-6679)

Allows the `score_reset` signal to be silenced during a `reset_score` operation.


efischer19:
Pairs with https://github.com/edx/edx-ora2/pull/984

PR #57, opened 2017-04-03
return serialized scores from get_scores()

jcdyer:
# [TNL-6697](https://openedx.atlassian.net/browse/TNL-6697)

This lets edx-platform use the created_at timestamp to estimate first_attempted for subsection grades.

## Reviewers

- [ ] @efischer19 

FYI @edx/educator-neem


submissions/api.py
```diff
@@ -626,7 +626,7 @@ def get_score(student_item):
 
 
 def get_scores(course_id, student_id):
-    """Return a dict mapping item_ids -> (points_earned, points_possible).
+    """Return a dict mapping item_ids -> [ScoreSerializer(ScoreSummary)].
```
efischer19: Since we're taking the `data` property in particular of a Serializer, could this more accurately be `Return a dict mapping item_ids -> JSON-style object representation`?

setup.py
```diff
@@ -33,7 +33,7 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='edx-submissions',
-    version='1.2.0',
+    version='2.0.0',
```
efischer19: I'm not really sure that this necessitates a 2.0.0 release as opposed to 1.3.0, but I don't actually have that strong of an opinion so carry on.

jcdyer: If we're using semantic versioning, this is a backwards incompatible api change: http://semver.org/, which does require a major version change.  I'm not sure how strict we are about using that.  I hear people say we use it, but we have a lot of code that just hangs out in v1.

efischer19: > I'm not sure how strict we are about using that.  I hear people say we use it, but we have a lot of code that just hangs out in v1.

You're adding more strictness to edx-submissions' "let's follow semver guidelines" process with this PR than has ever existed in this repo previously :rofl:.  I had been thinking that we could have added this in a backwards compatible way, but edx-ora2 is the only consumer so let's make it work exactly how we want.

jcdyer: Sure.  Half of our repos are languishing in pre-1.0 land despite being in heavy use in production.  I guess I just don't get the aversion to bumping the version number appropriately.  Given the way we work, we should have a lot more v13.2.1 level version numbers, and I think that's perfectly fine.

PR #58, opened 2017-07-10
Upgrade DRF requirement to v3.6.x

doctoryes:
As part of the Django 1.11 upgrade of edx-platform, the platform's required version of DRF needs to be upgraded. As this repo is a requirement of edx-platform and is pinned at a lower DRF version, the required version needs to be bumped here as well.

The platform will still use an edX fork of DRF - but that fork is v3.6.3 plus one patch.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,7 @@
 dogapi==1.2.1
 django-extensions==1.5.9
 django-model-utils==2.3.1
-# Use the same cherry-picked repo as edx-platform
-#djangorestframework>=3.1,<3.2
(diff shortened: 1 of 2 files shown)


doctoryes:
@efischer19 Want to take another look at this PR after I performed some minor changes? The more loosely bounded DRF requirement I had before did not work in full edx-platform testing.

efischer19:
Makes sense, I assume this one does work in that context then?

Still 👍 from me to merge, this looks logical.

doctoryes:
Yes - it does work in the edx-platform context now. I will merge - then tag this as 2.0.1

PR #61, opened 2017-07-20
Print DRF errors upon invalid serialization.

doctoryes:
https://openedx.atlassian.net/browse/LEARNER-1977

@iloveagent57 FYI


doctoryes:
@efischer19 @iloveagent57 I'm going to go ahead and merge this to get the edx-platform work started for this CAT-2 bug above. Please review it anyway.

iloveagent57:
I noticed earlier today that the `__version__` in `submissions/__init__.py` is stale, so if you have an opportunity to bump the version there, too... :)

iloveagent57:
`__version__` aside, this LGTM, too.

doctoryes:
We should simply delete one of those versions - probably using the __init__.py version is the way to go, reading/using that value in setup.py. I didn't see this in time to fix it my latest submissions PR. ☹️

PR #63, opened 2017-07-21
Fix exception printing and log upon error.

doctoryes:
Actually show the field errors upon the exception.

@efischer19 @iloveagent57 Please review.

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -89,6 +89,7 @@ class SubmissionRequestError(SubmissionError):
             if field_errors is not None
             else {}
         )
+        self.args += (self.field_errors,)
 
     def __repr__(self):
         """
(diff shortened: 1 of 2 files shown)


submissions/api.py
```diff
@@ -991,6 +992,11 @@ def _get_or_create_student_item(student_item_dict):
             student_item_serializer = StudentItemSerializer(
                 data=student_item_dict)
             if not student_item_serializer.is_valid():
+                logger.error(
+                    "Invalid StudentItemSerializer: {}".format(
+                        unicode(student_item_serializer.errors)
+                    )
```
doctoryes: Reviewers: While I logged the field_errors above, I didn't actually log the `student_item_dict`. Now we know from Splunk that the problem is an invalid course_id - but we don't know what the course_id value actually is. ☹️ 🙍‍♂️ 🙍  😦

PR #67, opened 2017-08-04
Efischer/edu 1090

efischer19:
# [EDUCATOR-1090](https://openedx.atlassian.net/browse/EDUCATOR-1090)

Requires the revert of #66, as this needs to jump the line and go out faster than https://github.com/edx/edx-platform/pull/15736. Without the revert, we have a version mismatch issue re: jsonfield.

FYI @sstack22 @jibsheet @MichaelRoytman

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -196,6 +196,20 @@ def create_submission(student_item_dict, answer, submitted_at=None, attempt_numb
         raise SubmissionInternalError(error_message)
 
 
+def _get_submission_model(uuid, read_replica=False):
+    """
+    Helper to retrieve a given Submission object from the database. Helper is needed to centralize logic that fixes
+    EDUCATOR-1090, because uuids are stored both with and without hyphens.
+    """
+    submission_qs = Submission.objects
+    if read_replica:
+        submission_qs = _use_read_replica(submission_qs)
+
+    query_regex = "^{}$|^{}$".format(uuid, uuid.replace("-",""))
+    submission = submission_qs.get(uuid__regex=query_regex)
+    return submission
+
+
 def get_submission(submission_uuid, read_replica=False):
     """Retrieves a single submission by uuid.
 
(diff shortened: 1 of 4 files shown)


submissions/tests/test_api.py
```diff
@@ -68,11 +68,7 @@ def test_get_submission_and_student(self):
         retrieved = api.get_submission_and_student(submission['uuid'])
         self.assertItemsEqual(submission, retrieved)
 
-        # Should raise an exception if uuid is malformed
-        with self.assertRaises(api.SubmissionInternalError):
-            api.get_submission_and_student(u'no such uuid')
-
-        # Should raise a different exception if the student item does not exist
```
efischer19: This fix is time-sensitive, so I'm not bothering to suss out the refactoring needed to ensure all our failures are the correct type. Malformed and correctly-formed-but-not-present values both return NotFound for now.

efischer19: I can address this in the next release, which is still in-flight but needs more time than we currently have.

submissions/api.py
```diff
@@ -196,6 +196,20 @@ def create_submission(student_item_dict, answer, submitted_at=None, attempt_numb
         raise SubmissionInternalError(error_message)
 
 
+def _get_submission_model(uuid, read_replica=False):
```
jibsheet: BTW - how do I get read_replica set to true in a future release?

submissions/api.py
```diff
@@ -196,6 +196,20 @@ def create_submission(student_item_dict, answer, submitted_at=None, attempt_numb
...
+    """
+    Helper to retrieve a given Submission object from the database. Helper is needed to centralize logic that fixes
+    EDUCATOR-1090, because uuids are stored both with and without hyphens.
+    """
+    submission_qs = Submission.objects
+    if read_replica:
+        submission_qs = _use_read_replica(submission_qs)
+
+    query_regex = "^{}$|^{}$".format(uuid, uuid.replace("-",""))
+    submission = submission_qs.get(uuid__regex=query_regex)
```
efischer19: Doing the regex forces string comparison on the database values, where uuid = A OR uuid = B will convert both A and B to the same (non-matching) formatted UUID hex value

PR #69, opened 2017-08-07
Return to normalcy

efischer19:
This reverts the revert I recently made, which was needed to get a hotfix out without blowing up requirements. Now that we look to be out of the woods there, this change can go back into master in anticipation of edx-ora2 2.0.0.

I'll also likely be adding some management command improvements to this PR, since I'm here.

### DIFF
--- a/submissions/management/commands/update_submissions_uuids.py
+++ b/submissions/management/commands/update_submissions_uuids.py
@@ -10,6 +10,7 @@ EDUCATOR-1090
 """
 
 import logging
+import time
 
 from django.core.management.base import BaseCommand
 from django.db import transaction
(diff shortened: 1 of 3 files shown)


efischer19:
@jibsheet I've updated the management command here while reverting my hotfix revert, no rush on the review as I've still got to sort out ORA acceptance test failures for the big release to land.

submissions/management/commands/update_submissions_uuids.py
```diff
@@ -24,21 +25,43 @@ class Command(BaseCommand):
...
+        """
+        Add arguments to the command parser.
+
+        Uses argparse syntax.  See documentation at
+        https://docs.python.org/3/library/argparse.html.
+        """
+        parser.add_argument(
+            '--start', '-s',
+            default=0,
+            help=u"0-based index at which to begin the update. Default 0."
```
jibsheet: Do you want to mention this is the id of the submissions table here?  Helping other folk understand what to feed it or what to target.

efischer19: Ah yeah, good call. I'll update.

PR #95, opened 2019-05-06
INCR-184: Fix TestCase.assertItemsEqual()

anthonywee:
assertItemsEqual() has been renamed to assertCountEqual()

https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertItemsEqual


anthonywee:
@edx/incr-reviewers

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5444](https://openedx.atlassian.net/browse/OSPR-5444) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #97, opened 2019-05-07
bump version number and enforce python 3 testing in travis

estute:
Bumping to version 3.0.0 as python3 testing is now officially supported. Will tag after merging


setup.py
```diff
@@ -58,6 +58,9 @@ def load_requirements(*requirements_paths):
         'License :: OSI Approved :: GNU Affero General Public License v3',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
+        'Programming Language :: Python 2.7',
+        'Programming Language :: Python 3.6',
+        'Programming Language :: Python 3.7',
```
jmbowman: Looks like you're missing the separator between Python and the version numbers.  Also, tools like `caniusepython3` and `requires.io` key off of the `Programming Language :: Python :: 3` classifier, so that should be included too.  Reference list: https://pypi.org/classifiers/

Up for adding classifiers for the tested Django versions while you're at it?  We'll want to do that anyway as we get ready for the Django 2.2 upgrade.

estute: Sure thing. Done.

PR #101, opened 2019-11-13
Django22 Support

Ayub-Khan:
-Updated tox
-Updated travis
-Updated django requirements
-Updated trove classifiers
-Version Bump
-test fixes
-Pip-tools for requirement management
-Added tox quality env
-Fixed quality
-Requirements constraints
-moved linting configs to setup.cfg
-added tox envs for drf versions
-toxs docs env


setup.py
```diff
@@ -55,21 +56,22 @@ def load_requirements(*requirements_paths):
...
-        'Framework :: Django :: 2.1',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: GNU Affero General Public License v3',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
```
awais786: Remove

.travis.yml
```diff
@@ -33,3 +73,4 @@ deploy:
     secure: BlYK3lplRoPiNfzqslxyReJZrC1QUT9Ls1A1xIxc3IHxKl9WEkqC2xO5vBvQDFKtMyooq8bGeN/gfksaZpRYqBmrO7hPITiz0FH0wPFdjoIikI4NTqKVxjhqUho0ZWmpxNXrX/iMd9Y2R4uZRXZGaWaTSRHfF9fPVlkbo66s7vw=
   on:
     tags: true
+    condition: "$TOXENV = django111"
```
awais786: Add python version.

submissions/admin.py
```diff
@@ -1,13 +1,13 @@
...
 from django.contrib import admin
-from django.core.urlresolvers import reverse
-from django.utils import html
+from django.urls import reverse
 
 from submissions.models import Score, ScoreSummary, StudentItem, Submission
 
 
-class StudentItemAdminMixin(object):
+class StudentItemAdminMixin():
```
jmbowman: This is a good improvement for Python-3-only code, but changes behavior on Python 2; it changes this from a [new-style class to an old-style class](https://docs.python.org/2/reference/datamodel.html#newstyle).

submissions/management/commands/tests/test_analyze_uploaded_file_sizes.py
```diff
@@ -33,7 +29,8 @@ def capture():
         out[1] = out[1].getvalue()
 
 
-class BaseMixin(object):
+class BaseMixin():
```
jmbowman: Same comment about class types here.

submissions/urls.py
```diff
@@ -1,11 +1,12 @@
+""" Submissions api urls. """
 from __future__ import absolute_import
 
-from django.conf.urls import patterns, url
```
jmbowman: Given that this function was removed in Django 1.10, I don't think this file is even in use (here or in edx-platform).  The code below is still slightly wrong, but it should be safe to just delete the whole file.

setup.py
```diff
@@ -55,21 +56,21 @@ def load_requirements(*requirements_paths):
...
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.6',
-        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.5',
+        'Framework :: Django',
+        'Framework :: Django :: 1.11',
+        'Framework :: Django :: 2.0',
+        'Framework :: Django :: 2.1',
+        'Framework :: Django :: 2.2',
```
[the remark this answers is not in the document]
Ayub-Khan: framework should be mentioned after language. but i can revert if you say so.

requirements/test.txt
```diff
@@ -26,48 +22,50 @@ factory-boy==2.12.0
 faker==2.0.4              # via factory-boy
 filelock==3.0.12          # via tox
 freezegun==0.3.12
-funcsigs==1.0.2           # via mock, pytest
-futures==3.3.0 ; python_version == "2.7"  # via isort
```
jmbowman: It looks like you ran `make upgrade` on Python 3, so it dropped all the dependencies needed for 2.7.  We should keep running that on the lowest version of Python we support to get as comprehensive of a dependency list as possible.

setup.py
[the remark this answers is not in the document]
jmbowman: I don't think I've ever seen anything stating a recommended order for the classifiers, but all the examples in the docs use alphabetical order and that's what we had here also.  Not a huge deal either way, but I'm curious if you've seen advice to the contrary somewhere.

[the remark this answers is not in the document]
Ayub-Khan: no, have not seen any thing. i have reverted the changes.

setup.py
```diff
@@ -55,21 +56,21 @@ def load_requirements(*requirements_paths):
...
     ],
     packages=['submissions', 'submissions.migrations'],
-    install_requires=load_requirements('requirements.txt', 'django-requirements.txt'),
-    tests_require=load_requirements('test-requirements.txt'),
+    install_requires=load_requirements('requirements/base.txt'),
+    tests_require=load_requirements('requirements/test.txt'),
```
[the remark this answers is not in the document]
Ayub-Khan: It seems tests fails when we use base.in and test.in files.  unable to ifnd files.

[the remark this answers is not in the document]
jmbowman: You need to add them to `MANIFEST.in` instead of the `.txt` files.

PR #102, opened 2019-11-14
Jkantor/mgmt

jansenk:
Create management command to gather and print data on assignment upload sizes

### DIFF
--- a/submissions/management/commands/analyze_uploaded_file_sizes.py
+++ b/submissions/management/commands/analyze_uploaded_file_sizes.py
@@ -32,6 +32,7 @@ EARLIEST_ALLOWED_DATE = datetime.date(2019, 11, 1)
 EARLIEST_ALLOWED_DATE_STR = EARLIEST_ALLOWED_DATE.strftime(DATE_FORMAT)
 HEADER = u'Course ID\tUsers with Uploaded Files\tTotal Uploaded Bytes\tAverage Upload per User'
 
+
 class Command(BaseCommand):
     """
     Example usage:
--- a/submissions/management/commands/analyze_uploaded_file_sizes.py
+++ b/submissions/management/commands/analyze_uploaded_file_sizes.py
@@ -9,8 +9,8 @@ Per course:
 This report is printed to stdout with a header and values as a tabbed stream.
 
 Arguments:
-    -start_date (optional): The most recent day to include in the report. Defaults to today.
-    -end_date (optional): The longest-ago day to include in the report.
+    -max_date (optional): The most recent day to include in the report. Defaults to today.
+    -min_date (optional): The longest-ago day to include in the report.
                           Defaults to <today - 30 days>.
                           The maximum value for this argument is 2019-11-01.
                           Before this day, file size data is unavailable.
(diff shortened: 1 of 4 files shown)


submissions/management/commands/analyze_uploaded_file_sizes.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def beginning_of_day(self, dt):
+        """ Given a datetime, return a datetime that represents the first second of that day """
+        return datetime.datetime.combine(dt, datetime.time.min)
+
+    def end_of_day(self, dt):
+        """ Given a datetime, return a datetime that represents the last second of that day """
+        return datetime.datetime.combine(dt, datetime.time.max)
+
+    def print_report(self, submission_data):
```
iloveagent57: Could you update this function name, or docstring (or both) to indicate that this is where the "group by" is effectively done?

submissions/management/commands/analyze_uploaded_file_sizes.py
```diff
@@ -0,0 +1,157 @@
...
+                course_bytes = 0
+                users = set()
+            user_submission_bytes = sum(answer['files_sizes'])
+            if user_submission_bytes:
+                users.add(student_id)
+                course_bytes += user_submission_bytes
+            prev_course_id = course_id
+
+        if prev_course_id:
+            self.print_row(prev_course_id, users, course_bytes)
```
iloveagent57: Or maybe turn this whole section into a helper function that returns lists of aggregated, per-course data, and let `print_report()` be in charge of only calling that function and printing the report?

submissions/management/commands/analyze_uploaded_file_sizes.py
```diff
@@ -0,0 +1,157 @@
...
+
+import datetime
+import json
+
+from django.core.management.base import BaseCommand, CommandError
+
+from submissions.models import Submission
+
+DATE_FORMAT = '%Y-%m-%d'
+LAST_DATE = datetime.date(2019, 11, 1)
```
iloveagent57: Nit: would be more clear as `EARLIEST_ALLOWED_DATE`.

submissions/management/commands/analyze_uploaded_file_sizes.py
```diff
@@ -0,0 +1,157 @@
...
+            '--start_date',
+            type=parse_date,
+            default=today,
+            help=u"The most recent day to include in the report. Defaults to today."
+        )
+        parser.add_argument(
+            '--end_date',
+            type=parse_date,
+            default=thirty_days_ago,
+            help=(u"The longest-ago day to include in the report. Defaults to <today - 30 days>. "
```
iloveagent57: This is the opposite of what I would expect `end-date` to mean. :) . Should we change to `min-date` and `max-date` instead?

submissions/management/commands/analyze_uploaded_file_sizes.py
```diff
@@ -0,0 +1,157 @@
...
+    def handle(self, start_date, end_date, *args, **options):
+        """
+        Analyze ORA file upload submissions and and print tab-limited report
+        """
+        print(u'Starting file upload submission report')
+        arg_echo_str = u'start_date = {}, end_date = {}'.format(
+            start_date.strftime(DATE_FORMAT),
+            end_date.strftime(DATE_FORMAT)
+        )
+        print(arg_echo_str)
```
iloveagent57: 👍  This is helpful

submissions/management/commands/analyze_uploaded_file_sizes.py
```diff
@@ -0,0 +1,157 @@
...
+        submission_data = self.load_data(start_date, end_date)
+        self.print_report(submission_data)
+
+    def validate_input_dates(self, start_date, end_date):
+        """
+        Checks the validity of start_date and end_date
+        Checks that the end date is not before LAST_DATE, and if it is, overrides it.
+        """
+        if start_date < end_date:
+            raise CommandError("End date must be before start date")
```
iloveagent57: This will all read much nicer if we switch to min/max date.

submissions/management/commands/analyze_uploaded_file_sizes.py
```diff
@@ -0,0 +1,157 @@
...
+        if datetime.date.today() < start_date:
+            print(u"Warning: start_date is in the future")
+        if end_date < LAST_DATE:
+            print(u'File size data is unavailable before 2019-11-01. Setting end_date to 2019-11-01')
```
iloveagent57: Nit: use format string with the `LAST_DATE` constant (or whatever you might rename it to).

PR #106, opened 2020-03-25
TeamSubmission model added.

Andytr1:
Jira: [EDUCATOR-4979](https://openedx.atlassian.net/browse/EDUCATOR-4979)
Update Submission models to support team based work. 
Add a TeamSubmission model and the supporting migrations.
Added unit tests for the pre_save signal.


submissions/models.py
```diff
@@ -107,6 +109,98 @@ def from_db_value(self, value, expression, connection, context=None):
...
 
 
+# Has this submission been soft-deleted? This allows instructors to reset student
+# state on an item, while preserving the previous value for potential analytics use.
+DELETED = u'D'
+ACTIVE = u'A'
+STATUS_CHOICES = (
+    (DELETED, u'Deleted'),
+    (ACTIVE, u'Active'),
+)
```
iloveagent57: Nit: since we're moving this, let's get rid of the `u` prefix before the string literals, since we're in python 3 now and don't need 'em.

submissions/models.py
```diff
@@ -107,6 +109,98 @@ def from_db_value(self, value, expression, connection, context=None):
...
+
+
+@python_2_unicode_compatible
+class TeamSubmission(TimeStampedModel):
+    """
+    A single response by a team for a given problem (ORA2) in a given course.
+    An abstraction layer over Submission used to for teams. Since we create a submission record for every team member,
+    there is a need to have a single point to connect the team to the workflows.
+    TeamSubmission is a many to many with Submission
```
iloveagent57: Nit: It's a 1-many, not many-many.

submissions/models.py
```diff
@@ -107,6 +109,98 @@ def from_db_value(self, value, expression, connection, context=None):
...
+
+    # doosNOT need it    's own index if the UNIQUE constraint detailed below exists
+    # (with course_id as first field)
+    course_id = models.CharField(max_length=255, null=False, db_index=True)
+
+    item_id = models.CharField(max_length=255, db_index=True)
+
+    team_id = models.CharField(max_length=255, db_index=True)
+
+    submitted_by = models.ForeignKey(User, on_delete=models.CASCADE)
```
iloveagent57: Hmm, what do you think of setting `null=True` for this field and then using `on_delete=models.SET_NULL`?

submissions/models.py
```diff
@@ -107,6 +109,98 @@ def from_db_value(self, value, expression, connection, context=None):
...
+    # Which attempt is this? Consecutive Submissions do not necessarily have
+    # increasing attempt_number entries -- e.g. re-scoring a buggy problem.
+    attempt_number = models.PositiveIntegerField()
+
+    # submitted_at is separate from created_at to support re-scoring and other
+    # processes that might create Submission objects for past user actions.
+    submitted_at = models.DateTimeField(default=now, db_index=True)
+
+    # doosNOT need it    's own index if the UNIQUE constraint detailed below exists
+    # (with course_id as first field)
```
iloveagent57: Nit: you can delete this comment, since we decided we couldn't have that unique constraint.

submissions/models.py
```diff
@@ -107,6 +109,98 @@ def from_db_value(self, value, expression, connection, context=None):
...
+    course_id = models.CharField(max_length=255, null=False, db_index=True)
+
+    item_id = models.CharField(max_length=255, db_index=True)
+
+    team_id = models.CharField(max_length=255, db_index=True)
```
iloveagent57: Nit: might as well set `null=False` on these fields, too.

submissions/models.py
```diff
@@ -107,6 +109,98 @@ def from_db_value(self, value, expression, connection, context=None):
...
+            created=self.created,
+            modified=self.modified,
+        ))
+
+    def __str__(self):
+        return u"Team Submission {}".format(self.uuid)
+
+    class Meta:
+        app_label = u"submissions"
+        ordering = ["-submitted_at", "-id"]
```
iloveagent57: Nit: we don't need these `u`s, either.

submissions/models.py
```diff
@@ -107,6 +109,98 @@ def from_db_value(self, value, expression, connection, context=None):
...
+
+
+@receiver(pre_save, sender=TeamSubmission)
+def validate_only_one_submission_per_team(sender, **kwargs):
+    """
+    Ensures that there is only one submission per team.
+    """
```
iloveagent57: Nit: only one _active_ submission per team.

submissions/models.py
```diff
@@ -143,20 +237,14 @@ class Submission(models.Model):
...
-    # state on an item, while preserving the previous value for potential analytics use.
-    DELETED = u'D'
-    ACTIVE = u'A'
-    STATUS_CHOICES = (
-        (DELETED, u'Deleted'),
-        (ACTIVE, u'Active'),
-    )
     status = models.CharField(max_length=1, choices=STATUS_CHOICES, default=ACTIVE)
 
+    team_submission = models.ForeignKey(TeamSubmission, null=True, db_index=True, on_delete=models.SET_NULL)
```
iloveagent57: Nit: let's give this a `related_name` attribute like `related_name=submissions`.

submissions/tests/test_models.py
```diff
@@ -166,3 +170,52 @@ def test_highest_score_hidden(self):
...
+        with pytest.raises(DuplicateTeamSubmissionsError):
+            TeamSubmission.objects.create(
+                submitted_by=user,
+                team_id='t1',
+                course_id='c1',
+                item_id='i1',
+                attempt_number=1
+            )
+
+    @classmethod
```
iloveagent57: Nit: you can make this a `staticmethod` instead, since you're not using `cls`.  And then you'll be able to do `self.create_user()` in your test functions.

submissions/tests/test_models.py
```diff
@@ -166,3 +170,52 @@ def test_highest_score_hidden(self):
...
         self.assertEqual(highest.points_possible, 2)
+
+
+class TestTeamSubmission(TestCase):
+    """
+    Test the TeamSubmission class
+    """
+    def test_create_team_submission(self):
+        user = TestTeamSubmission.create_user('user1')
+        ts = TeamSubmission.objects.create(
```
iloveagent57: I'd recommend doing your object creation in a `setUpTestData` method: https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.TestCase.setUpTestData

That way, any objects you create will be rolled back once this test class is finished, and we won't have to worry about other tests in the platform being thrown off if they do a count of users in the `User` table or something like that.

submissions/models.py
```diff
@@ -107,6 +109,98 @@ def from_db_value(self, value, expression, connection, context=None):
...
+    def __str__(self):
+        return u"Team Submission {}".format(self.uuid)
+
+    class Meta:
+        app_label = u"submissions"
+        ordering = ["-submitted_at", "-id"]
+
+
+@receiver(pre_save, sender=TeamSubmission)
+def validate_only_one_submission_per_team(sender, **kwargs):
```
iloveagent57: Good idea 👍

submissions/models.py
Andytr1: I like that suggestion better, since we at least preserve a team submission record.

submissions/tests/test_models.py
Andytr1: Python must hate me. For some reason my `@class method setupTestData` doesn't get called :(

iloveagent57: You're defining it like below, right? With a capital `U` in `Up`?
```python
    @classmethod
    def setUpTestData(cls):
        # Set up data for the whole TestCase
```

Andytr1: I need to find a Homer Simpson 'DOH'! emoticon.

submissions/tests/test_models.py
```diff
@@ -166,3 +178,56 @@ def test_highest_score_hidden(self):
...
+
+class TestTeamSubmission(TestCase):
+    """
+    Test the TeamSubmission class
+    """
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = TestTeamSubmission.create_user('user1')
+        super(TestTeamSubmission, cls).setUpTestData()
```
iloveagent57: Nit: python 3 allows you to simply call `super()` without arguments.

submissions/tests/test_models.py
```diff
@@ -166,3 +178,56 @@ def test_highest_score_hidden(self):
...
+
+
+class TestTeamSubmission(TestCase):
+    """
+    Test the TeamSubmission class
+    """
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = TestTeamSubmission.create_user('user1')
```
iloveagent57: Nit: `cls.user = cls.create_user('user1')`

submissions/tests/test_models.py
```diff
@@ -166,3 +178,56 @@ def test_highest_score_hidden(self):
...
+        super(TestTeamSubmission, cls).setUpTestData()
+
+    def test_create_team_submission(self):
+        ts = TeamSubmission.objects.create(
```
iloveagent57: Nit: you could create class-level variables to store these `TeamSubmission` objects in inside of `setUpTestData`, so that they get cleaned up after all the tests in the class run.

submissions/tests/test_models.py
Andytr1: What is the process/location of bumping the python version?

iloveagent57: https://github.com/edx/edx-submissions/blob/master/submissions/__init__.py#L2

Andytr1: Apologies. Long day.

PR #110, opened 2020-04-16
EDUCATOR-4983: Create the Team Submissions python API

jansenk:
[EDUCATOR-4983](https://openedx.atlassian.net/browse/EDUCATOR-4983)
@edx/masters-devs-gta 

changes:
- added `team_api.py`, a new api layer for creating/interacting with team submissions.
- added several static methods to `TeamSubmission`. Much of the equivalent code for `Submission`s are in `api.py` but i thought it made sense to keep most lookup utility functions as model methods.

refactors:
- took alex's suggested change from the draft and changed the way that `api.py`'s `create_submission` method looked up older active submissions using `first()`
- moved all custom errors into `errors.py`


submissions/models.py
```diff
@@ -160,6 +162,92 @@ def get_queryset(self):
...
+
+        """
+        model_query_params = dict(course_id=course_id, item_id=item_id, team_id=team_id)
+        query_params_string = "course_id={course_id} item_id={item_id} team_id={team_id}".format(**model_query_params)
+        try:
+            # In the equivalent non-teams api call, we're filtering on student item and then getting first(),
+            # which will get us the most recent active submission due to the sort order of the model.
+            # However, for this model, we have a uniqueness constraint (non-db, as a signal handler)
+            # that means we can only ever have one submission per team per assignment. I don't fully understand
+            # the logic behind the non-teams api, but this shouldn't have to do that filter.
```
jansenk, line 206: This last comment shouldn't make it in, but let's please discuss.

submissions/team_api.py
```diff
@@ -0,0 +1,240 @@
+"""
+Public interface for team submissions API.
+"""
+
+# We're going to need common things like `SubmissionError`
```
jansenk, line 5: remove

submissions/team_api.py
```diff
@@ -0,0 +1,240 @@
...
+    This api function:
+      1. Creates a `TeamSubmission` record, and
+      2. Creates `Submission` records for every member of the team by calling api.create_submission()
+
+    This means that the ORA `SubmissionMixin` must first collect all of the files of the submitting user
+    and the team into the `answer` dict.
+
+    Returns:
+        dict: A representation of the created TeamSubmission, with the following keys:
+          'team_submission_uuid' Is the `uuid` field of the created `TeamSubmission`.
```
jansenk, line 38: change comment to reflect that it's student_item_dict

submissions/team_api.py
```diff
@@ -0,0 +1,240 @@
...
+            'answer': u'The answer is 42.',
+            'submission_uuids': ['alice-uuid', 'bob-uuid', 'carol-uuid', 'tim-uuid'],
+        }
+    """
+
+    # I have no clue what to do with attempt_number. The normal API checks if there are any duplicate submissions and
+    # incrememnts attempt_number, but we prevent duplicate team submissions, so I can't copy that logic flow.
+    # I thought maybe we should look for deleted submissions and incrememnt based of of that, but looking at
+    # production data I can't find a single case where that happens.
+    # For now ... I'm just gonna default it to 1?
```
jansenk, line 90: This comment is slightly too casual but if we don't come up with a better idea, I think leaving something like this in here is a good idea

submissions/serializers.py
```diff
@@ -86,6 +117,7 @@ class Meta:
             'submitted_at',
             'created_at',
             'answer',
+            'team_submission'
```
jansenk, line 120: looking at this now, we may want to change this to team_submission_uuid

submissions/team_api.py
jansenk: Also, to expand on this comment, what I did find in production was that there would be a deleted submission, and then an active submission with the same attempt_number

submissions/tests/test_team_api.py
```diff
@@ -0,0 +1,401 @@
...
+            COURSE_ID,
+            ITEM_1_ID,
+            TEAM_1_ID,
+            self.user_1.id,
+            self.user_ids,
+            ANSWER
+        )
+
+    @freeze_time("2020-04-10 12:00:01", tz_offset=-4)
+    def test_create_submission_for_team(self):
```
nsprenkle, line 116: This unit test is beautiful 💯

submissions/tests/test_team_api.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def test_create_submission_for_team_existing_deleted_team_submission(self):
+        """
+        Test for calling create_submission_for_team with an existing deleted team submission
+        for the same team course and item.
+        """
+        team_submission_1 = self.make_team_submission(status=DELETED)
+        team_submission_2 = self._call_create_submission_for_team_with_default_args()
+        self.assertEqual(team_submission_1.attempt_number, 1)
+        self.assertEqual(team_submission_2['attempt_number'], 1)
```
nsprenkle, line 170: Yeah... the attempt numbering is weird for teams...

submissions/tests/test_team_api.py
```diff
@@ -0,0 +1,401 @@
...
+                ITEM_1_ID,
+                TEAM_1_ID,
+                self.user_1.id,
+                self.user_ids,
+                ANSWER,
+                attempt_number=-1
+            )
+
+    def test_create_submission_for_team_existing_individual_submission(self):
+        """
```
nsprenkle, line 188: Can you explain what the behavior here is?

submissions/tests/test_models.py
```diff
@@ -226,3 +223,88 @@ def create_team_submission(user, team_id='team1', course_id='c1', item_id='i1',
             item_id=item_id,
             attempt_number=attempt_number
         )
+
+    def test_create_team_submission(self):
+        # force evaluation of __str__ to ensure there are no issues with the class, since there
+        # isn't much specific to assert.
+        self.assertNotEqual(self.default_submission.__str__, None)
```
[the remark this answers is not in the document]
jansenk: I'm not sure why this is showing up as new? I definitely didn't write this. Did I maybe copy it from somewhere else? Either way, I think this whole test can just go away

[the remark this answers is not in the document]
jansenk: Oh, looks like I just moved it. Can still be removed, it even says itself it's not really doing much

submissions/tests/test_team_api.py
jansenk: I don't think this will ever happen, but in case we ever get into a weird data state I wanted this behavior explicitly demonstrated in a test.

For an individual submission, if we call create_submisison when there is an existing submission, we will increment the attempt_number of the most recent submission. For teams, though, we pass the `attempt_number` from `create_team_submission` to `create_submission` so that all of the numbers match. I don't know that it's .... important? It's just kinda weird so I wanted to be explicit.

jansenk: I'll expand this comment

submissions/models.py
```diff
@@ -160,6 +162,92 @@ def get_queryset(self):
...
+        Raises:
+            - TeamSubmissionNotFoundError if there is no matching team submission
+            - TeamSubmissionInternalError if there is some other error looking up the team submission.
+        """
+        try:
+            return TeamSubmission.objects.prefetch_related('submissions').get(uuid=team_submission_uuid)
+        except TeamSubmission.DoesNotExist:
+            logger.error("Team Submission {} not found.".format(team_submission_uuid))
+            raise TeamSubmissionNotFoundError(
+                u"No team submission matching uuid {}".format(team_submission_uuid)
```
Andytr1, line 179: nit: u prefix no longer needed in python 3

submissions/team_api.py
```diff
@@ -0,0 +1,246 @@
...
+        'course_id': course_id,
+        'item_id': item_id,
+        'team_id': team_id,
+        'submitted_by': submitting_user_id,
+        'attempt_number': attempt_number,
+    }
+    if submitted_at:
+        model_kwargs["submitted_at"] = submitted_at
+
+    try:
```
Andytr1, line 103: When googling @transaction.atomic - I found the [doc page](https://docs.djangoproject.com/en/2.2/topics/db/transactions/) with the following quote (mid page) : 

> Avoid catching exceptions inside atomic!

Basically they recommend to double handle exceptions. Should we try and follow that pattern?

jansenk: I think that they're saying to follow that pattern when you're using the context manager inside of a try:catch. Either way, I tried it, and it doesn't seem like it really... fits?

plus, the example they give fits the following form: 
```
@transaction.atomic
def f():
   create_parent()
   try:
     with transaction.atomic():
        create_children()
    catch databasseerror:
       blah()
```

jansenk: we're more or less doing the same thing here, but we aren't trying to catch any eceptions when creating the children, so they'll jut propogate up. I think it should be okay?

jansenk: @nsprenkle do you have aany thoughts on this?

Andytr1: Doing some more code scanning in our code base(platform) we have examples of what you are doing. I think we should be good to go.

PR #111, opened 2020-04-20
BOM-1544: Added python 3.8 in test matrix

iamsobanjaved:
**JIRA:** [BOM-1544](https://openedx.atlassian.net/browse/BOM-1544)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,27 +3,14 @@ dist: latest
 
 python:
   - 3.5
+  - 3.8
 
 env:
-  - TOXENV=django111-drf39
-  - TOXENV=django111-drf310
-  - TOXENV=django111-drflatest
-  - TOXENV=django20-drf39
-  - TOXENV=django20-drf310
-  - TOXENV=django20-drflatest
-  - TOXENV=django21-drf39
-  - TOXENV=django21-drf310
-  - TOXENV=django21-drflatest
   - TOXENV=django22-drf39
   - TOXENV=django22-drf310
   - TOXENV=django22-drflatest
-
-matrix:
-  include:
-    - python: 3.5
-      env: TOXENV=quality
-    - python: 3.5
-      env: TOXENV=docs
+  - TOXENV=quality
+  - TOXENV=docs
 
 install:
   - pip install -r requirements/test.txt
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -3,17 +3,9 @@ dist: latest
...
 env:
-  - TOXENV=django111-drf39
-  - TOXENV=django111-drf310
-  - TOXENV=django111-drflatest
-  - TOXENV=django20-drf39
-  - TOXENV=django20-drf310
-  - TOXENV=django20-drflatest
-  - TOXENV=django21-drf39
-  - TOXENV=django21-drf310
-  - TOXENV=django21-drflatest
```
UsamaSadiq, line 16: You can also move the `quality` and `docs` envs from `matrix` to `envs` and remove the `matrix` section altogether.

iamsobanjaved: @UsamaSadiq IMO there is no need to run quality and docs test on python 3.5 as well as on python 3.8

tox.ini
```diff
@@ -1,14 +1,12 @@
 [tox]
-envlist = py35-django{111,20,21,22}-drf{39,310,latest}
+envlist = py{35,38}-django{22,30}-drf{39,310,latest}
```
morenol, line 2: Django 3.0 with Python3.5 is an invalid combination

tox.ini
```diff
@@ -1,14 +1,12 @@
 [tox]
-envlist = py35-django{111,20,21,22}-drf{39,310,latest}
+envlist = py{35,38}-django22-drf{39,310,latest}, py38-django30
```
jmbowman, line 2: Won't we still want to test multiple DRF versions under Django 3.0?

iamsobanjaved: There isn't any released version of DRF which supports Django 3.0. Github's master branch of DRF has support for Django 3.0. Should we test django 3.0 against that?

jmbowman: Or just use `latest`, and we'll see if they've released a new version with Django 3.0 support by the time we get around to actually testing it.  We're not actually adding Django 3.0 test configurations in Travis yet.

iamsobanjaved: Added drf latest with django 3.0

PR #114, opened 2020-04-27
Return answer as part of submissions team API

Andytr1:
We need to return answer as part of create_team_submission.

### DIFF
--- a/submissions/serializers.py
+++ b/submissions/serializers.py
@@ -57,6 +57,30 @@ class TeamSubmissionSerializer(serializers.ModelSerializer):
     submitted_at = DateTimeField(format=None, required=False)
     created_at = DateTimeField(source='created', format=None, required=False)
     attempt_number = IntegerField(min_value=0)
+    # Prevent Django Rest Framework from converting the answer (dict or str)
+    # to a string.
+    # answer is not a part of TeamSubmission model. We populate it externally.
+    answer = serializers.SerializerMethodField()
+
+    def get_answer(self, obj):  # pylint: disable=unused-argument
+        answer = self.context.get("answer")
+        return answer
+
+    def validate_answer(self, value):
+        """
+        Check that the answer is JSON-serializable and not too long.
+        """
+        # Check that the answer is JSON-serializable
+        try:
+            serialized = json.dumps(value)
+        except (ValueError, TypeError):
+            raise serializers.ValidationError("Answer value must be JSON-serializable")
+
+        # Check the length of the serialized representation
+        if len(serialized) > Submission.MAXSIZE:
+            raise serializers.ValidationError("Maximum answer size exceeded.")
+
+        return value
 
     class Meta:
         model = TeamSubmission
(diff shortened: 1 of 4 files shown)


submissions/team_api.py
```diff
@@ -101,7 +101,7 @@ def create_submission_for_team(
         model_kwargs["submitted_at"] = submitted_at
 
     try:
-        team_submission_serializer = TeamSubmissionSerializer(data=model_kwargs)
+        team_submission_serializer = TeamSubmissionSerializer(data=model_kwargs, context={'answer': answer})
```
nsprenkle, line 104: I think, for consistency, this should follow the `api.py` version,  where answer is part of the original `model_kwargs` unless there's a specific reason to break the pattern. It should result in fewer changes.

submissions/serializers.py
```diff
@@ -57,6 +57,29 @@ class TeamSubmissionSerializer(serializers.ModelSerializer):
...
+    # Prevent Django Rest Framework from converting the answer (dict or str)
+    # to a string.
+    answer = serializers.SerializerMethodField()
+
+    def get_answer(self, obj):  # pylint: disable=unused-argument
+        answer = self.context.get("answer")
+        return answer
+
+    def validate_answer(self, value):
+        """
```
jansenk, line 69: I was gonna say that we already do this check at the individual submisisons level, but i think I actually like putting this here.

submissions/serializers.py
```diff
@@ -57,6 +57,29 @@ class TeamSubmissionSerializer(serializers.ModelSerializer):
     submitted_at = DateTimeField(format=None, required=False)
     created_at = DateTimeField(source='created', format=None, required=False)
     attempt_number = IntegerField(min_value=0)
+    # Prevent Django Rest Framework from converting the answer (dict or str)
+    # to a string.
```
jansenk, line 61: Maybe toss a comment on here saying that although this is included in the serializer that we save to a model, this value isn't a model field

submissions/team_api.py
jansenk: there's no answer field on TeamSubmission, so i think that would raise an exception

submissions/serializers.py
```diff
@@ -57,6 +57,29 @@ class TeamSubmissionSerializer(serializers.ModelSerializer):
...
+    answer = serializers.SerializerMethodField()
+
+    def get_answer(self, obj):  # pylint: disable=unused-argument
```
jansenk, line 64: I don't think there's anything wrong with passing in the context dict, but if we want to avoid that I think we could also do 

`return obj.submissions.first().answer`

submissions/team_api.py
nsprenkle: Ooh, my bad. I was thinking that `TeamSubmission` inherited from `Submission`. Carry on!

submissions/serializers.py
Andytr1: Cool. Didn't realize that was an option either. I will use context if no major objections.

PR #118, opened 2020-05-18
EDUCATOR-5059: Teams Submissions are using usernames for submission student_id rather than anonymous user ids

jansenk:
[EDUCATOR-5059](https://openedx.atlassian.net/browse/EDUCATOR-5059)

Submissions didn't really have much functional change that needed to be done, since it more or less just takes what it's given and stores it, but I wanted to clarify some stuff in the docstring, and change the tests to be reflective of how we actually want the api to be used.

@edx/masters-devs-gta

### DIFF
--- a/submissions/__init__.py
+++ b/submissions/__init__.py
@@ -1,2 +1,2 @@
 """ API for creating submissions and scores. """
-__version__ = '3.1.6'
+__version__ = '3.1.7'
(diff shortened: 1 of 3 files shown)


submissions/__init__.py
```diff
@@ -1,2 +1,2 @@
 """ API for creating submissions and scores. """
-__version__ = '3.1.6'
+__version__ = '3.1.7'
```
Andytr1, line 2: Does it make sense to increment version if we are just adding doc string/fixing unit tests?

jansenk: I think so, my understanding is that it goes <major>.<minor>.<build> so really any code change should bump that last number

PR #121, opened 2020-06-05
Return a team submission given regular submission uuid associated with it

Andytr1:
In ORA2, there are cases where we need to retrieve a team submission given regular submission uuid associated with it. This API method does it.
[Jira](https://openedx.atlassian.net/browse/EDUCATOR-5089)

@edx/masters-devs-gta

### DIFF
--- a/submissions/team_api.py
+++ b/submissions/team_api.py
@@ -7,7 +7,12 @@ import logging
 from django.db import DatabaseError, transaction
 
 from submissions import api as _api
-from submissions.errors import SubmissionInternalError, TeamSubmissionInternalError, TeamSubmissionRequestError
+from submissions.errors import (
+    SubmissionInternalError,
+    TeamSubmissionInternalError,
+    TeamSubmissionNotFoundError,
+    TeamSubmissionRequestError
+)
 from submissions.models import DELETED, TeamSubmission
 from submissions.serializers import TeamSubmissionSerializer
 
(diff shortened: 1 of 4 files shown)


Andytr1:
jenkins run all

submissions/team_api.py
```diff
@@ -187,6 +192,21 @@ def get_team_submission(team_submission_uuid):
...
+
+    Raises:
+        - TeamSubmissionNotFoundError when no such team submission exists.
+        - TeamSubmissionInternalError if there is some other error looking up the team submission.
+    """
+    team_submission_query = TeamSubmission.objects.filter(submissions__uuid=individual_submission_uuid).all()
+    if len([team_submission_query]) < 1:
+        raise TeamSubmissionNotFoundError
+
+    return TeamSubmissionSerializer(team_submission_query[0]).data
```
nsprenkle, line 207: Proposing a cleaner version using `first()`:

```suggestion
    team_submission = TeamSubmission.objects.filter(submissions__uuid=individual_submission_uuid).first()
    if not team_submission:
        raise TeamSubmissionNotFoundError

    return TeamSubmissionSerializer(team_submission).data
```

submissions/tests/test_team_api.py
```diff
@@ -237,6 +237,19 @@ def test_get_team_submission(self):
             TeamSubmissionSerializer(team_submission_model).data
         )
 
+    def test_get_team_submission_from_individual_submission(self):
```
nsprenkle, line 240: Is there a case where we would expect the exception case to happen? If so, it might also use a test case.

Andytr1: No specific case. This is for defensive purposes.

submissions/team_api.py
Andytr1: Nice. I like it!

submissions/tests/test_team_api.py
```diff
@@ -237,6 +237,19 @@ def test_get_team_submission(self):
...
+        """
+        Test that calling team_api.get_team_submission_from_individual_submission returns the expected team submission
+        """
+        team_submission_model = self._make_team_submission(create_submissions=True)
+        regular_submission_uuid = [team_submission_model.submissions.all()][0][0].uuid
```
jansenk, line 245: what's the second `[0]` there for? we have a list of lists of submissions? 
is it because we're putting the brackets around the all() and then having a list with one element which is the list of submissions? 

could we just do `list(team_submission_model.submissions.all())[0]`? or even better, `team_submission_model.submissions.first().uuid`?

submissions/tests/test_team_api.py
jansenk: i'd probably agree with nathan, probably best for completeness' sake to put a teensy test in that hits that use case

jansenk:
oh also you have to bump version

submissions/tests/test_team_api.py
Andytr1: good catch. writing code Friday afternoon does interesting things to one's perspective.

PR #123, opened 2020-06-17
Set score as an atomic transaction

nsprenkle:
Jira: [EDUCATOR-5017](https://openedx.atlassian.net/browse/EDUCATOR-5017)
FYI: @edx/masters-devs-gta

### DIFF
--- a/submissions/api.py
+++ b/submissions/api.py
@@ -12,7 +12,7 @@ from uuid import UUID
 import six
 from django.conf import settings
 from django.core.cache import cache
-from django.db import DatabaseError, IntegrityError
+from django.db import DatabaseError, IntegrityError, transaction
 
 # SubmissionError imported so that code importing this api has access
 from submissions.errors import (  # pylint: disable=unused-import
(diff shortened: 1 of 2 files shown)


submissions/api.py
```diff
@@ -845,29 +845,28 @@ def set_score(submission_uuid, points_earned, points_possible,
...
+            if annotation_creator is not None:
+                score_annotation = ScoreAnnotation(
+                    score=score_model,
+                    creator=annotation_creator,
+                    annotation_type=annotation_type,
+                    reason=annotation_reason
+                )
+                score_annotation.save()
+            # Send a signal out to any listeners who are waiting for scoring events.
+            score_set.send(
```
jansenk, line 861: I don't think that the signal sending needs to be a part of the atomic block

jansenk:
technically, version? idk this is such an incredibly small change

submissions/api.py
nsprenkle: Ah, that makes sense, but still part of the `try` block.

PR #124, opened 2020-07-07
[BD-18] Replace nose with pytest

tehreem-sadat:
**Description:**

- Replace nose with pytest.
- Update packages with "make upgrade"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,6 @@ htmlcov/
 .tox/
 .coverage
 .cache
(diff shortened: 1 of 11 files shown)


edx-webhook:
Thanks for the pull request, @tehreem-sadat! I've created [BLENDED-404](https://openedx.atlassian.net/browse/BLENDED-404) to keep track of it in Jira.

When this pull request is ready, tag your edX technical lead.

PR #125, opened 2020-07-15
BOM-1914 : Fix Django 3.0 deprecation warnings

mzulqarnain1:


submissions/models.py
```diff
@@ -105,7 +105,7 @@ class UpdatedJSONField(JSONField):
     # because we're using jsonfield2 version 3.0.3 and it has a known issue while performing
     # select_related queries, it has been fixed in version 3.1.0 but python3.5 support is also
     # dropped in that version, for now we've put the fix here
-    def from_db_value(self, value, expression, connection, context=None):
+    def from_db_value(self, value, expression, connection):  # pylint: disable=arguments-differ
```
mzulqarnain1, line 108: pylint warning was disabled because it is a false positive. 
`W0221: Parameters differ from overridden 'from_db_value' method (arguments-differ)`

PR #128, opened 2020-08-14
EDUCATOR-5210 

Andytr1:
Don't create new submissions for a learner that has an existing team submission.

[Jira](https://openedx.atlassian.net/browse/EDUCATOR-5210)

@edx/masters-devs-gta

### DIFF
--- a/submissions/__init__.py
+++ b/submissions/__init__.py
@@ -1,2 +1,2 @@
 """ API for creating submissions and scores. """
-__version__ = '3.1.12'
+__version__ = '3.2.1'
--- a/submissions/team_api.py
+++ b/submissions/team_api.py
@@ -13,7 +13,7 @@ from submissions.errors import (
     TeamSubmissionNotFoundError,
     TeamSubmissionRequestError
 )
-from submissions.models import DELETED, TeamSubmission, Submission
+from submissions.models import DELETED, Submission, TeamSubmission
 from submissions.serializers import TeamSubmissionSerializer
 
 logger = logging.getLogger(__name__)
--- a/submissions/team_api.py
+++ b/submissions/team_api.py
@@ -13,7 +13,7 @@ from submissions.errors import (
     TeamSubmissionNotFoundError,
     TeamSubmissionRequestError
 )
-from submissions.models import DELETED, TeamSubmission
+from submissions.models import DELETED, TeamSubmission, Submission
 from submissions.serializers import TeamSubmissionSerializer
 
 logger = logging.getLogger(__name__)
@@ -34,7 +34,8 @@ def create_submission_for_team(
     """
     This api function:
       1. Creates a `TeamSubmission` record, and
-      2. Creates `Submission` records for every member of the team by calling api.create_submission()
+      2. Creates `Submission` records for ONLY those team members that don't have any past submissions
+         by calling api.create_submission()
 
     This means that the ORA `SubmissionMixin` must first collect all of the files of the submitting user
     and the team into the `answer` dict.
@@ -137,7 +138,20 @@ def create_submission_for_team(
         'item_id': item_id,
         'item_type': item_type
     }
+
+    existing_student_ds_team = [submission.student_item.student_id for submission in Submission.objects.filter(
+        team_submission__isnull=False
+    ).exclude(
+        team_submission__team_id=team_id
+    ).filter(
+        student_item__student_id__in=team_member_ids,
+        student_item__course_id=course_id,
+        student_item__item_id=item_id
+    ).select_related('team_submission').select_related('student_item')]
+
     for team_member_id in team_member_ids:
+        if team_member_id in existing_student_ds_team:
+            continue
         team_member_student_item_dict = dict(base_student_item_dict)
         team_member_student_item_dict['student_id'] = team_member_id
         _api.create_submission(
--- a/submissions/team_api.py
+++ b/submissions/team_api.py
@@ -35,7 +35,6 @@ def create_submission_for_team(
     This api function:
       1. Creates a `TeamSubmission` record, and
       2. Creates `Submission` records for ONLY those team members that don't have any past submissions
-         by calling api.create_submission()
 
     This means that the ORA `SubmissionMixin` must first collect all of the files of the submitting user
     and the team into the `answer` dict.
@@ -139,7 +138,7 @@ def create_submission_for_team(
         'item_type': item_type
     }
 
-    existing_student_ds_team = [submission.student_item.student_id for submission in Submission.objects.filter(
+    students_with_team_submissions = [submission.student_item.student_id for submission in Submission.objects.filter(
         team_submission__isnull=False
     ).exclude(
         team_submission__team_id=team_id
@@ -150,7 +149,7 @@ def create_submission_for_team(
     ).select_related('team_submission').select_related('student_item')]
 
     for team_member_id in team_member_ids:
-        if team_member_id in existing_student_ds_team:
+        if team_member_id in students_with_team_submissions:
             continue
         team_member_student_item_dict = dict(base_student_item_dict)
         team_member_student_item_dict['student_id'] = team_member_id
--- a/submissions/team_api.py
+++ b/submissions/team_api.py
@@ -13,7 +13,7 @@ from submissions.errors import (
     TeamSubmissionNotFoundError,
     TeamSubmissionRequestError
 )
-from submissions.models import DELETED, Submission, TeamSubmission
+from submissions.models import DELETED, StudentItem, TeamSubmission
 from submissions.serializers import TeamSubmissionSerializer
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


submissions/__init__.py
```diff
@@ -1,2 +1,2 @@
 """ API for creating submissions and scores. """
-__version__ = '3.1.12'
+__version__ = '3.2.1'
```
nsprenkle, line 2: Confirm numbering before merging (i.e. if this merges before the bump to `3.2.1`, make this the `3.2.0` version.

submissions/team_api.py
```diff
@@ -34,7 +34,8 @@ def create_submission_for_team(
     """
     This api function:
       1. Creates a `TeamSubmission` record, and
-      2. Creates `Submission` records for every member of the team by calling api.create_submission()
+      2. Creates `Submission` records for ONLY those team members that don't have any past submissions
```
nsprenkle, line 37: Nit: I'd leave out the comment about it calling `api.create_submission()`, seems like an unnecessary detail which ties high-level docstring too much to implementation detail.

submissions/team_api.py
```diff
@@ -137,7 +138,20 @@ def create_submission_for_team(
         'item_id': item_id,
         'item_type': item_type
     }
+
+    existing_student_ds_team = [submission.student_item.student_id for submission in Submission.objects.filter(
```
nsprenkle, line 142: What is a `student_ds_team`?

submissions/tests/test_team_api.py
```diff
@@ -194,24 +194,58 @@ def test_create_submission_for_team_invalid_attempt(self):
...
-        an existing active submission for the item.
-
-        For normal Submissions, if a submission exists and we create another one, the second
-        will increment the first's attempt_number. However, for team submissions, we currently
-        pass the attempt_number from team_api.create_submission to api.create_submission, so
-        all created individual submissions will have the same attempt_number
+        an existing active submission for the item in the same course.
+        This can happen is a user was on a team that submitted and then reassigned to a team
+        that didn't have a submission and is submitting.
+        Expected outcome: submission is created because the the previous submission was not team based
```
nsprenkle, line 200: Grammar/typo fixes. Also, an item is by definition in the same course so you can remove that:
```suggestion
        an existing active submission for the item.
        This can happen if a user was on a team that submitted and then is reassigned to a team
        that didn't have a submission and is submitting.
        Expected outcome: submission is created because the the previous submission was not team based
```

submissions/tests/test_team_api.py
```diff
@@ -194,24 +194,58 @@ def test_create_submission_for_team_invalid_attempt(self):
...
             'student_item'
-        ).filter(
-            uuid__in=team_submission_data['submission_uuids']
-        ).all()
-        for submission in submissions:
-            self.assertEqual(submission.attempt_number, 1)
+        ).filter(uuid__in=team_submission_data['submission_uuids']).all()]
+        self.assertIn(user_3_item.student_id, submission_student_ids)
+
+    def test_create_submission_for_team_existing_individual_submission_different_course(self):
```
nsprenkle, line 211: I don't think you need this test: teams/items by definition can't exist across courses, can they?

submissions/tests/test_team_api.py
```diff
@@ -194,24 +194,58 @@ def test_create_submission_for_team_invalid_attempt(self):
...
+        SubmissionFactory.create(student_item=user_3_item)
+        team_submission_data = self._call_create_submission_for_team_with_default_args()
+        self.assertEqual(team_submission_data['attempt_number'], 1)
+
+        submission_student_ids = [sub.student_item.student_id for sub in Submission.objects.select_related(
+            'student_item'
+        ).filter(uuid__in=team_submission_data['submission_uuids']).all()]
+        self.assertIn(user_3_item.student_id, submission_student_ids)
+
+    def test_create_submission_for_team_after_reassignment(self):
```
nsprenkle, line 229: This is the real logic we want to test but it's unclear based on how it's written. Why do we expect 4 submission items instead of another number? Since the test functions don't explicitly tell you what the teams/submissions look like, the comments need to be more clear.

submissions/__init__.py
Andytr1: Yep. I made this version in anticipation of Jansen's pr being merged before mine.

submissions/team_api.py
Andytr1: Thank you. That should be ids.

submissions/team_api.py
```diff
@@ -137,7 +137,20 @@ def create_submission_for_team(
         'item_id': item_id,
         'item_type': item_type
     }
+
+    students_with_team_submissions = [submission.student_item.student_id for submission in Submission.objects.filter(
```
jansenk, line 141: I think this can be replaced? You're querying on Submission, but you're really lookup up StudentItems. 

```
[student_item.student_id for student_item in StudentItem.objects.filter(
        submission_set__team_submission__isnull=False
    ).exclude(
        submission_set___team_submission__team_id=team_id
    ).filter(
        student_id__in=team_member_ids,
        course_id=course_id,
        item_id=item_id
    )]

```

jansenk: Also, you have `select_related('team_submission')` but it doesn't look like you're navigating through that

submissions/team_api.py
```diff
@@ -137,7 +137,20 @@ def create_submission_for_team(
...
+    ).exclude(
+        team_submission__team_id=team_id
+    ).filter(
+        student_item__student_id__in=team_member_ids,
+        student_item__course_id=course_id,
+        student_item__item_id=item_id
+    ).select_related('team_submission').select_related('student_item')]
+
     for team_member_id in team_member_ids:
+        if team_member_id in students_with_team_submissions:
```
jansenk, line 152: should probably make `students_with_team_submissions` a set then

submissions/tests/test_team_api.py
```diff
@@ -195,23 +195,42 @@ def test_create_submission_for_team_existing_individual_submission(self):
...
         Test for calling create_submission_for_team when a user somehow already has
         an existing active submission for the item.
-
-        For normal Submissions, if a submission exists and we create another one, the second
-        will increment the first's attempt_number. However, for team submissions, we currently
-        pass the attempt_number from team_api.create_submission to api.create_submission, so
-        all created individual submissions will have the same attempt_number
+        This can happen if a user was on a team that submitted and then is reassigned to a team
+        that didn't have a submission and is submitting.
+        Expected outcome: submission is created because the previous submission was not team based
```
jansenk, line 200: This is probably good to test just for error behavior, but let's make it clear that this would require staff to release an ORA as an individual ORA, a user to submit, then for staff to change to ORA to team based and then student to submit again.

jansenk: Haha wait I wrote this test lmao

jansenk: This is kinda a silly test in retrospect, but hey, good on ya for not just deleting it :p

submissions/tests/test_team_api.py
```diff
@@ -195,23 +195,42 @@ def test_create_submission_for_team_existing_individual_submission(self):
...
             'student_item'
-        ).filter(
-            uuid__in=team_submission_data['submission_uuids']
-        ).all()
-        for submission in submissions:
-            self.assertEqual(submission.attempt_number, 1)
+        ).filter(uuid__in=team_submission_data['submission_uuids']).all()]
+        self.assertIn(user_3_item.student_id, submission_student_ids)
+
+    def test_create_submission_for_team_after_reassignment(self):
```
jansenk, line 211: This is a kinda odd case, just since it's not expected behavior, as in we shouldn't really expect a case where every member of a team already has a submission. If anything, this should be an error state where we don't create the TeamSubmission and raise an exception

submissions/tests/test_team_api.py
```diff
@@ -195,23 +195,42 @@ def test_create_submission_for_team_existing_individual_submission(self):
...
+        """
```
jansenk, line 212: How about a test where you do the same setup and then call

```
team_api.create_submission_for_team(
            COURSE_ID,
            ITEM_1_ID,
            TEAM_2_ID,
            self.user_1.id,
           [self.anonymous_user_id_map[self.user_1], 'anotheruser-1', 'anotheruser-2', 'anotheruser-3']
            ANSWER
        )

```

or something like that, then you can assert that the TeamSubmission was successfully created for the second team, but that there wasn't a submission created for user_1?

submissions/tests/test_team_api.py
Andytr1: Ask and you shall receive :)

submissions/tests/test_team_api.py
```diff
@@ -223,14 +223,15 @@ def test_create_submission_for_team_after_reassignment(self):
...
         )
         ids = [sub.student_item.student_id for sub in Submission.objects.select_related('student_item').all()]
-        # We simulated reassignment by using a different team id in the call to create_submission. Therefore, only
-        # the 4 submissions from the first call should exist, as we would not create submissions for learners if
-        # they already had team submissions
-        self.assertEqual(len(self.student_ids), len(ids))
+        # We simulated reassignment by using a different team id in the call to create_submission. Therefore, 6
+        # submissions should exist: 4 from the first call and 2 from the second call. We would not createa
+        # submission for user_1 in the second call because she already has a submission from the first call.
+        self.assertEqual(6, len(ids))
```
jansenk, line 234: looks good. :+1: any chance you could add in an assertion that loops through the student ids in this test and asserts that each user has exactly one submission?

jansenk:
Also I merged my PR so you can rebase off master now

submissions/tests/test_team_api.py
```diff
@@ -208,23 +208,46 @@ def test_create_submission_for_team_existing_individual_submission(self):
...
+            ANSWER
+        )
+        ids = [sub.student_item.student_id for sub in Submission.objects.select_related('student_item').all()]
+        # We simulated reassignment by using a different team id in the call to create_submission. Therefore, 6
+        # submissions should exist: 4 from the first call and 2 from the second call. We would not createa
+        # submission for user_1 in the second call because she already has a submission from the first call.
+        self.assertEqual(6, len(ids))
+        # this assert checks that there is one and only one (no duplicate - which would indicate a double submission)
+        # student id
+        self.assertEqual(len(ids), len(set(ids)))
```
jansenk, line 250: 👍

PR #129, opened 2020-08-19
add get_submissions_from_other_teams

muselesscreator:
#What does this do?
Provides a new api method from the submission team_api that returns a list of objects for each student in a given team with a submission to a given item made from another team.
Returned values include the student_id as well as the team_id as there are different consumers for this data.

FIY: @edx/masters-devs-gta

### DIFF
--- a/submissions/team_api.py
+++ b/submissions/team_api.py
@@ -138,15 +138,14 @@ def create_submission_for_team(
         'item_type': item_type
     }
 
-    students_with_team_submissions = {student_item.student_id for student_item in StudentItem.objects.filter(
-        submission__team_submission__isnull=False
-    ).exclude(
-        submission__team_submission__team_id=team_id
-    ).filter(
-        student_id__in=team_member_ids,
-        course_id=course_id,
-        item_id=item_id
-    )}
+    students_with_team_submissions = [
+        submission['student_id'] for submission in get_submissions_from_other_teams(
+            team_member_ids,
+            team_id,
+            item_id,
+            course_id
+        )
+    ]
     for team_member_id in team_member_ids:
         if team_member_id in students_with_team_submissions:
             continue
@@ -190,6 +189,40 @@ def _log_team_submission(team_submission_data):
             submitted_by=team_submission_data["submitted_by"],
         )
     )
+
+
+def get_submissions_from_other_teams(
+    team_member_ids,
+    team_id,
+    item_id,
+    course_id,
+):
+    """
+    This api function returns a list of objects for students on this team that have submitted
+    a response to the given item under another team.
+
+    Parameters:
+        - team_member_ids (list of str): a list of the anonymous user ids associated with all members of the team
+        - team_id (str): the team_id for the team for which we are making the submission
+        - item_id (str): the item id for this team submission
+        - course_id (str): the course id for this team submission
+
+    Returns:
+        set: { 'student_id', 'team_id' }
+    """
+    items = StudentItem.objects.filter(
+        submission__team_submission__isnull=False
+    ).exclude(
+        submission__team_submission__team_id=team_id
+    ).filter(
+        student_id__in=team_member_ids,
+        course_id=course_id,
+        item_id=item_id
+    ).values("student_id", "submission__team_submission__team_id")
+    return [{
+        'student_id': item['student_id'],
+        'team_id': item['submission__team_submission__team_id']
+    } for item in items]
 
 
 def get_team_submission(team_submission_uuid):
--- a/submissions/tests/test_team_api.py
+++ b/submissions/tests/test_team_api.py
@@ -261,6 +261,44 @@ class TestTeamSubmissionsApi(TestCase):
         self.assertEqual(TeamSubmission.objects.count(), 0)
         self.assertEqual(Submission.objects.count(), 0)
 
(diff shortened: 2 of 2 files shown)


submissions/team_api.py
```diff
@@ -138,15 +138,14 @@ def create_submission_for_team(
...
-    students_with_team_submissions = {student_item.student_id for student_item in StudentItem.objects.filter(
-        submission__team_submission__isnull=False
-    ).exclude(
-        submission__team_submission__team_id=team_id
-    ).filter(
-        student_id__in=team_member_ids,
-        course_id=course_id,
-        item_id=item_id
-    )}
+    students_with_team_submissions = [
```
Andytr1, line 141: Why did you change from a set to a list?

submissions/team_api.py
```diff
@@ -192,6 +191,40 @@ def _log_team_submission(team_submission_data):
     )
 
 
+def get_submissions_from_other_teams(
```
Andytr1, line 194: The name doesn't "jive" with what the method actually does. Trying to think of a better one.

submissions/team_api.py
muselesscreator: Cant set dictionaries. (unless there's a trick I forgot... sorry, rusty python chops)

muselesscreator: I could add a line to enforce uniqueness by student_id after this query?

Andytr1: You could just change from `[` and  `]` to `{`  and `}` to initialize a set instead of a list.

submissions/team_api.py
jansenk: I agree, this name sounds like it's getting the Submission  objects

get_teammates_with_submissions_from_other_teams() ?

submissions/team_api.py
```diff
@@ -192,6 +191,40 @@ def _log_team_submission(team_submission_data):
...
+    team_member_ids,
+    team_id,
+    item_id,
+    course_id,
+):
+    """
+    This api function returns a list of objects for students on this team that have submitted
```
jansenk, line 201: list of dicts

submissions/team_api.py
```diff
@@ -192,6 +191,40 @@ def _log_team_submission(team_submission_data):
...
+    a response to the given item under another team.
+
+    Parameters:
+        - team_member_ids (list of str): a list of the anonymous user ids associated with all members of the team
+        - team_id (str): the team_id for the team for which we are making the submission
+        - item_id (str): the item id for this team submission
+        - course_id (str): the course id for this team submission
```
jansenk, line 208:
```suggestion
        - team_id (str): the team_id of the team for which we are making the query
        - item_id (str): the item id for which we are checking for submissions
        - course_id (str): the course id for which we are checking for submissions
```

submissions/team_api.py
```diff
@@ -192,6 +191,40 @@ def _log_team_submission(team_submission_data):
...
+
+    Returns:
+        set: { 'student_id', 'team_id' }
```
jansenk, line 211: returns a list of dicts, not a set

submissions/team_api.py
```diff
@@ -192,6 +191,40 @@ def _log_team_submission(team_submission_data):
...
+    """
+    items = StudentItem.objects.filter(
+        submission__team_submission__isnull=False
+    ).exclude(
+        submission__team_submission__team_id=team_id
+    ).filter(
+        student_id__in=team_member_ids,
+        course_id=course_id,
+        item_id=item_id
+    ).values("student_id", "submission__team_submission__team_id")
```
jansenk, line 221: Nice! does this make only one query?

submissions/tests/test_team_api.py
```diff
@@ -261,6 +261,45 @@ def test_create_submission_for_team_error_creating_individual_submission(self, m
...
+            status=None,
+            create_submissions=True
+        )
+        # Check against TEAM_2 with 2 additional user IDs added (that don't have a submission)
+        team_ids = [
+            self.anonymous_user_id_map[student] for student in [
+                self.user_1, self.user_2, self.user_3, self.user_4
+            ]
+        ] + ['55555555555555', '666666666666666666']
+        external_submissions = team_api.get_submissions_from_other_teams(
```
jansenk, line 280: would you mind tossing a `with self.assertNumQueries(1):` call around this line?

submissions/tests/test_team_api.py
```diff
@@ -261,6 +261,45 @@ def test_create_submission_for_team_error_creating_individual_submission(self, m
         self.assertEqual(TeamSubmission.objects.count(), 0)
         self.assertEqual(Submission.objects.count(), 0)
 
+    def test_get_submissions_from_other_teams(self):
```
jansenk, line 264: Good test

submissions/team_api.py
```diff
@@ -192,6 +191,40 @@ def _log_team_submission(team_submission_data):
     )
 
 
+def get_submissions_from_other_teams(
+    team_member_ids,
```
jansenk, line 195: this is like, the tiniest nit, but could you change the order of params to course, item, team? it's how we order it pretty mush everywhere else in this file, since it's sorta in descending """size""" order

Andytr1:
Don't forget to increase the version!

PR #136, opened 2021-02-18
Replace jsonfield2 with jsonfield

iamsobanjaved:
We are going to replace jsonfield2 with jsonfield as the earlier one has archived and merged back in the latter one.

**JIRA:** https://openedx.atlassian.net/browse/BOM-2413


submissions/management/commands/analyze_uploaded_file_sizes.py
```diff
@@ -111,7 +111,7 @@ def load_data(self, min_date, max_date):
         return Submission.objects.filter(
             created_at__range=(min_datetime, max_datetime),
             student_item__item_type='openassessment',
-            answer__contains='files_sizes":['
+            answer__contains='files_sizes": ['
```
iamsobanjaved, line 114: With the latest version of `jsonfield`, this query is now space sensitive

UsamaSadiq (approved):

iamsobanjaved:
We had to merge all jsonfield PRs in edX packages at once because this PR will bring latest version of jsonfield in platform as well which needs migration in few other packages as well.

iamsobanjaved:
Hi @edx/masters-devs-gta, can I get a review on this PR

submissions/models.py
```diff
@@ -98,20 +97,6 @@ class Meta:
         )
 
 
-class UpdatedJSONField(JSONField):
-    """
-    Class inherits JSONFiled from jsonfiled2 and provides overridden implementation
-    """
-    # We need to this custom class to represent JSONField instead of standard JSONField
-    # because we're using jsonfield2 version 3.0.3 and it has a known issue while performing
-    # select_related queries, it has been fixed in version 3.1.0 but python3.5 support is also
```
iamsobanjaved, line 107: As we are upgrading to latest, the fix is arrived so don't need this anymore

submissions/api.py
```diff
@@ -236,6 +236,11 @@ def get_submission(submission_uuid, read_replica=False):
         return cached_submission_data
 
     try:
+        # In previous versions of jsonfield, an error was raised whenever invalid json value was spotted,
+        # in version 3.1.0, this is changed and now a RuntimeWarning is emitted, so catching that warning
+        # to keep the endpoint response format consistent to previous version
+        warnings.filterwarnings('error')
+
```
iamsobanjaved, line 243: Changelog: https://github.com/rpkilby/jsonfield/blob/master/CHANGES.rst#v310-02222020

Commit that is changing error to warning: https://github.com/rpkilby/jsonfield/commit/c2f8780266ee32c25e2a3c3d727fba0e19d1bfb8

submissions/api.py
```diff
@@ -244,11 +249,19 @@ def get_submission(submission_uuid, read_replica=False):
...
         ) from error
-    except Exception as exc:
+    except (Exception, RuntimeWarning) as exc:
         # Something very unexpected has just happened (like DB misconfig)
+        # or an invalid json value is found in JSONField (submission.answer)
         err_msg = f"Could not get submission due to error: {exc}"
         logger.exception(err_msg)
+
+        # Switching filterwarnings back to its default behaviour
+        warnings.filterwarnings('default')
```
jmbowman, line 259: You shouldn't need this here since the `finally` block should be executed whether there was an exception or not.

submissions/migrations/0002_auto_20210223_0748.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('submissions', '0001_squashed_0005_CreateTeamModel'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='submission',
+            name='answer',
+            field=jsonfield.fields.JSONField(blank=True, db_column='raw_answer'),
```
jmbowman, line 17: It seems odd to both be editing the previous migration and adding this one.  Can the parameters in the previous migration be updated so this one isn't needed?  I don't think there are any actual database changes.

jmbowman (commented):
This came up in discussion on https://github.com/edx/upgrades/issues/34 ; is this the last blocker to getting the switch rolled out in edx-platform?  In addition to the points below, we probably need a rebase here to get updated requirement files and the correct new version to use.

iamsobanjaved:
@jmbowman There are a few other PRs that are blocked in the review cycle, those are listed below.

1. https://github.com/edx/xblock-lti-consumer/pull/182
2. https://github.com/edx/edx-ora2/pull/1595

PR #141, opened 2021-10-01
add get_student_ids_by_submission_uuid and get_team_ids_by_team_submission_uuid

jansenk:
Add two new API functions to enable lookups for the ORA Enhanced Staff Grader

get_student_ids_by_submission_uuid: given a course id and a list of submission uuids, return a dict mapping input submission_uuid to the associated student_id

get_team_ids_by_team_submission_uuid: given a list of team submission uuids, return a dict mapping input team submission uuid to team id

get_team_ids_by_team_submission_uuid is not currently used, and is not currently tested. I will be creating a ticket for this implementation
 - [ ] make get_team_ids_by_team_submission_uuid ticket

@edx/masters-devs-gta


jansenk:
> Should this be put into a feature branch or are we safe to merge here? (Assuming this is safe but want to confirm)

Should be safe to just merge

PR #144, opened 2021-11-10
Standardize version number placement

aht007:
Description:
Standardize the placement of version number

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,10 @@
 #!/usr/bin/env python
 
 import os
+import re
 import sys
 
 from setuptools import find_packages, setup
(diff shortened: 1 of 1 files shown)


mraarif (approved):

nsprenkle:
Just a note (maybe to @mattcarter) that the `version` field will have to be initialized as part of the merge.

aht007:
@nsprenkle that's not the case. The version field is already present in the `init` file and we are just changing the way how we fetch it in the setup.py file. We don't need to initialize anything.

iamsobanjaved (approved):

PR #145, opened 2021-11-12
refactor: Make team submissions optional in Django admin

nsprenkle:
**What changed?** For a "submission", change the related "team submission" field to optional. 

Jira: [AU-45](https://openedx.atlassian.net/browse/AU-45)

From dealing with a separate CR ticket I learned that you can't save changes to non-team ORAs in Django Admin (not that we frequently do this, but is a workflow for soft-deleting messed up submissions). Attempting to save gives you a validation error and blocks you from saving changes. This change should fix that.

**Testing instructions**

- [x] Load new code, `make install-local-submissions`
- [x] Log into Django admin, e.g. `http://{lms}/admin/`
- [x] Open an individual submission, `http://{lms}/admin/submissions/submission/`
- [x] Verify that it can be modified and saved with an empty team submission field

FYI @edx/masters-devs-gta


mattcarter (approved):
tested - works as described. Approved

PR #147, opened 2021-12-03
fix: attempt to optimize get_team_submission_student_ids

jansenk:
The old get_team_submission_student_ids query would take around 40 seconds. This takes 3 seconds, which is still not good, but it's better.


mattcarter (approved):
LGTM 👍

PR #150, opened 2021-12-18
fix:get_teammates_with_submissions_from_other_teams returns cancelled submissions

jansenk:
get_teammates_with_submissions_from_other_teams will currently return studentitem entries associated with deleted submissions.

Fix that, and also change the lookup model to Submission for query efficency's sake. The StudentItem query required some subqueries in its filtering, this is a very basic JOIN.

### DIFF
--- a/submissions/team_api.py
+++ b/submissions/team_api.py
@@ -13,7 +13,7 @@ from submissions.errors import (
     TeamSubmissionNotFoundError,
     TeamSubmissionRequestError
 )
-from submissions.models import DELETED, StudentItem, TeamSubmission
+from submissions.models import DELETED, Submission, TeamSubmission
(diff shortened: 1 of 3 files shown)


submissions/team_api.py
```diff
@@ -234,19 +234,20 @@ def get_teammates_with_submissions_from_other_teams(
...
     ).exclude(
-        submission__team_submission__team_id=team_id
-    ).filter(
-        student_id__in=team_member_ids,
-        course_id=course_id,
-        item_id=item_id
-    ).values("student_id", "submission__team_submission__team_id")
+        team_submission__isnull=True,
+    ).exclude(
+        team_submission__team_id=team_id,
```
jansenk, line 244: we don't need to actually specify we're excluding DELETED because the submissions manager handles that for us.

PR #153, opened 2022-01-13
style: fix pylint errors introduced by version bumps

matthugs:


matthugs:
codecov is failing because the refactors I’ve performed increased the number of lines in some places where test coverage doesn’t exist, but the changes were refactors so I don't think codecov being green is a blocker for this.

muselesscreator (approved):
LGTM

PR #157, opened 2022-02-22
fix: bug in get_team_ids_by_team_submission_uuid

nsprenkle:
**TL;DR** fix lookup in `get_team_ids_by_team_submission_uuid`

As part of ESG team support ticket ([AU-493](https://openedx.atlassian.net/browse/AU-493)) I ran into this bug. The delicious irony: we have an outstanding ticket to write tests ([AU-343](https://openedx.atlassian.net/browse/AU-343)) which would have caught this.

**NOTE** writing tests is going to wait until AU-343 for expediency.

### DIFF
--- a/submissions/__init__.py
+++ b/submissions/__init__.py
@@ -1,2 +1,2 @@
 """ API for creating submissions and scores. """
-__version__ = '3.5.0'
+__version__ = '3.5.1'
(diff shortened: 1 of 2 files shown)


jansenk (approved):
Does this API function not have any tests? Do we have a ticket for writing those?

nsprenkle:
> 

No tests. Yes we have a ticket. See PR description.

PR #168, opened 2022-06-27
build: add long_description to setup

jansenk:
Setuptools was bumped and now we need this

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,7 @@ setup(
     version=VERSION,
     author='edX',
     description='An API for creating submissions and scores.',
+    long_description="An API for creating and scoring submissions for the Open edX platform",
     url='http://github.com/edx/edx-submissions.git',
     license='AGPL',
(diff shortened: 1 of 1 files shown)


jansenk:
Not bumping version #, i'll just delete the tag and recreate

setup.py
```diff
@@ -52,6 +52,7 @@ def load_requirements(*requirements_paths):
     version=VERSION,
     author='edX',
     description='An API for creating submissions and scores.',
+    long_description="An API for creating and scoring submissions for the open edX platform",
```
nsprenkle, line 55: Nit: This is the correct branding: 

```suggestion
    long_description="An API for creating and scoring submissions for the Open edX platform",
```

nsprenkle (approved):
:shipit:

jansenk:
ugh github suggestion auto-commit default commit messages aren't CC-compliant 🙄

PR #173, opened 2022-08-18
Jkantor/ensure ascii

jansenk:
This fixes a bug where we are unable to store 4-byte UTF-8 unicode characters due to the encoding on this table. Rather than rebuilding the entire table with a different encoding, we thought the more straightforward fix was to simply json-encode characters before saving in the DB

I've tested this locally but I will be monitoring this release and will be testing on stage before the prod rollout. It does seem to work cleanly on devstack, so fingers crossed everything is nice

### DIFF
--- a/submissions/__init__.py
+++ b/submissions/__init__.py
@@ -1,2 +1,2 @@
 """ API for creating submissions and scores. """
-__version__ = '3.5.2'
+__version__ = '3.5.3'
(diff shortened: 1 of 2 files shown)


leangseu-edx (approved):
LGTM

nsprenkle (approved):

PR #208, opened 2023-08-07
feanil/update rtd docs

feanil:
This repo has a ReadTheDocs project that is currently broken, in making fixes for an upcoming RTD config deprecation, I noticed there were other issues with this repo.  Rather than dropping this from RTD, I'm fixing forward so that this repo has docs that are in compliance with our latest standards and that will not have issues with the RTD config deprecation late next month.

- docs: Add a ReadTheDocs config.
- docs: Move to the new sphinx book theme.
- docs: Update to be able to pull in docstrings.
- docs: Update docstrings to not throw warnings.
- chore: Run `make upgrade`

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,22 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required: the version of this file's schema.
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/source/conf.py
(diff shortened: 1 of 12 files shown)


kdmccormick:
Everything looks good, but could you add some context for these changes? e.g., what's the desired end state we're going for?

Looking at https://github.com/openedx/edx-submissions/tree/master/docs/source, this repo contains no RST docs. So, is the point of this to be able to publish docstrings on docs.openedx.org or elsewhere?

feanil:
@kdmccormick ah, thanks, yeas added some context to the PR description.  It kind of snowballed from dealing with a deprecation on the RTD side.

PR #227, opened 2024-02-26
feat: adding python3.12 support.

awais786:


.github/workflows/ci.yml
```diff
@@ -14,10 +14,13 @@ jobs:
...
       matrix:
         os: [ubuntu-20.04]
-        python-version: ['3.8']
+        python-version: ['3.8', '3.12']
         toxenv: [
-            quality, docs, django32-drf314, django42-drf314
+            quality, docs, django42-drf314
         ]
+        exclude:
+          - python-version: '3.12'
```
UsamaSadiq, line 22: quality checks for `python 3.12` require some fixes which don't work in `quality-pyton3.8` so we can't run both at the same time. 
This exclusion can be removed once the 3.8 support is completely removed after upgrading dependencies with python 3.12.

feanil (approved):

tox.ini
```diff
@@ -23,7 +22,8 @@ commands =
     make test_quality
 
 [testenv:docs]
-deps = 
+deps =
+    setuptools
```
UsamaSadiq, line 26: This shouldn't be needed for Python 3.11 support as `setuptools` is removed in `Python 3.12`.

Makefile
```diff
@@ -37,20 +37,34 @@ dev_requirements: ## Install Dev Requirements
...
+$(COMMON_CONSTRAINTS_TXT):
+	wget -O "$(@)" https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt || touch "$(@)"
+	echo "$(COMMON_CONSTRAINTS_TEMP_COMMENT)" | cat - $(@) > temp && mv temp $(@)
+
+export CUSTOM_COMPILE_COMMAND = make upgrade
+upgrade: piptools $(COMMON_CONSTRAINTS_TXT)	## update the requirements/*.txt files with the latest packages satisfying requirements/*.in
 	# Make sure to compile files after any other files they include!
 	pip-compile --upgrade --allow-unsafe --rebuild -o requirements/pip.txt requirements/pip.in
-	pip-compile --upgrade --verbose --rebuild -o requirements/pip-tools.txt requirements/pip-tools.in
+	pip-compile --upgrade --allow-unsafe --verbose --rebuild -o requirements/pip-tools.txt requirements/pip-tools.in
```
UsamaSadiq, line 57: ``--allow-unsafe`` shouldn't be needed for `Python 3.11` support.

PR #273, opened 2024-11-25
docs: Add PII annotations

bmtcril:
Per OEP-30 these annotations belong with their models, I'm just moving them from the edx-platform safelist to their correct homes.


pomegranited (approved):
:+1: Thanks @bmtcril :)

- [x] <strike>I tested this</strike> docs update only
- [x] I read through the code and checked the annotated models for PII
- [x] Updates documentation

PR #283, opened 2025-02-05
FC-73 XQueue add fields to submission model (Feature: ExternalGraderDetail - Gradual Migration from Xqueue System)

leoaulasneo98:
# FC-73 Feature: ExternalGraderDetail - Gradual Migration from Xqueue System

**⚠️ Important:** This PR is backward compatible and can be safely merged independently. While it was originally designed to work with submission structure changes in edx-platform, we've since changed the approach to create a new function instead of modifying existing ones. In fact, it's recommended to merge this PR before the related changes in edx-platform repository.

## Description
This pull request introduces the `ExternalGraderDetail` model as part of a gradual migration of the Open edX submission queueing system (xqueue) towards a more robust and maintainable submission processing system.

## Motivation 
The current xqueue system requires a strategic update to:
- Provide seamless transition for existing services (xwatcher)
- Enhance submission state management  
- Optimize processing and tracking of submission tasks

## Model Enhancements
- Introduced `ExternalGraderDetail` with:
 - Explicit state management
 - Consolidated timestamp tracking (`status_time`)
 - Atomic state transitions
 - Support for dynamic field assignment

## New Fields Added
- `queue_name`: Identifies processing queue
- `status`: Current submission queue state
- `status_time`: Unified timestamp for state tracking
- `pullkey`: Submission processing tracking key
- `grader_file_name`: Grading file metadata
- `grader_reply`: Grader response storage
- `points_possible`: Maximum possible points

## Queue Processing Improvements
- Introduced `SubmissionQueueManager` for queue query processing
- Implemented state transition validation
- Optimized database indexes for queue operations

## State Management
- Defined `STATUS_CHOICES`
- Implemented `VALID_TRANSITIONS`
- Created transaction-safe `update_status()` method

## Error Handling  
- Added custom `SubmissionQueueEmptyError`
- Improved field validation
- Safer dictionary access using `external_grader_detail.get()`

## Testing Strategy
Comprehensive test coverage includes:
- Basic model functionality
- State transition validation
- Error handling
- Integration with existing submission system

## Planned Developments
- Integrate with new edX submission services to meet xwatcher requirements
- Develop robust documentation as the implementation evolves
- Prepare for potential future enhancements in file handling and processing

## Open edX Compliance
Adheres to Open edX standards:
- Modular and extensible design
- Comprehensive error handling
- Extensive test coverage

## Performance Considerations
- Optimized database indexing
- Efficient state transition mechanism  
- Minimal processing overhead

[...]

PR #286, opened 2025-02-24
FC-73 Xqu add submission file (Feature: Submission File System - Enhanced File Management for Graded Submissions)

leoaulasneo98:
# FC-73 Feature: Submission File System - Enhanced File Management for Graded Submissions

## Description

This pull request implements a comprehensive file management system for submissions through the introduction of the `SubmissionFile` model and its associated `SubmissionFileManager`. This enhancement supports proper storage, retrieval, and processing of files attached to submissions within the Open edX ecosystem.

## Motivation

The existing submission system requires enhanced file handling capabilities to:
* Provide seamless transition for existing services (xwatcher)
* Establish reliable file storage with appropriate metadata
* Integrate seamlessly with the new submission queue architecture

Previously, the XQueue server managed files for submissions that required document attachments. However, this was implemented inefficiently using plain text fields (s3_keys and s3_urls) without proper structure or validation. This approach led to inconsistent file handling, potential security issues, and maintenance challenges as the system scaled.

### Model Enhancements
* Introduced `SubmissionFile` model with:
  * Secure file storage mechanisms
  * Proper association with submission queue records
  * Standardized URL generation for grader access
  * Original filename preservation

### File Processing System
* Developed `SubmissionFileManager` with capabilities for:
  * Processing multiple file formats (bytes, file objects)
  * Robust error handling for file processing issues
  * Consistent URL generation for xqueue compatibility
  * Efficient retrieval methods for external grader integration

### Error Handling
* Implemented comprehensive error handling for:
  * Invalid file objects
  * IO and OS exceptions during file reading
  * Unicode decoding errors
  * Non-standard file formats

### File Processing
* Support for multiple input types including:
  * Raw bytes
  * Django ContentFile objects
  * UploadedFile instances
  * Custom file-like objects with read() methods

### Integration Points
* Seamless connection with `ExternalGraderDetail` for complete submission workflow
* Compatible interface for existing xqueue consumers

## Testing Strategy
Extensive test coverage includes:
* Various file input types
* Complete processing workflow verification
* Edge case handling for invalid files
* Error handling for exceptional conditions
* File retrieval in grader-compatible format

## Open edX Compliance
This implementation adheres to Open edX standards through:
* Modular design with clear separation of concerns
* Comprehensive test coverage of edge cases
* Well-documented interfaces
* Consistent error handling patterns

## Performance Considerations
* Efficient file storage with minimal overhead
* Optimized database access patterns
* Careful handling of file objects to prevent memory issues

**BREAKING CHANGES**: None. Designed for full compatibility with existing systems.


angonz:
Review should only start after PR #283 is merged.

docs/source/decisions/0002-add-submission-file-model.rst
angonz, line 1: Update author of cange
Add link to first ADR
Summarize and simplify.

submissions/models.py
```diff
@@ -693,3 +696,102 @@ def update_status(self, new_status):
...
+        self.submission_queue = submission_queue
+
+    def process_files(self, files_dict):
+        """
+        Process uploaded files.
+        Returns URLs in xqueue compatible format.
+        """
+        files_urls = {}
+        for filename, file_obj in files_dict.items():
+            if not (isinstance(file_obj, (bytes, ContentFile, SimpleUploadedFile)) or hasattr(file_obj, 'read')):
```
angonz, line 760: Why check for `hasattr(file_obj, 'read')`?

submissions/models.py
```diff
@@ -693,3 +696,102 @@ def update_status(self, new_status):
...
+                logger.warning(f"Invalid file object type for {filename}")
+                continue
+
+            if hasattr(file_obj, 'read'):
```
angonz, line 764: Please explain this workflow

angonz (approved):
Ok on my side. Please open for Axim's review.

submissions/models.py
```diff
@@ -693,3 +702,136 @@ def update_status(self, new_status):
...
+        instance.external_grader.queue_name,
+        f"{instance.uid}"
+    )
+
+
+class SubmissionFile(models.Model):
+    """
+    Model to handle files associated with submissions
+    """
+    uid = models.UUIDField(default=uuid4, editable=False)  # legacy S3 key
```
ormsbee, line 723:
```suggestion
    uuid = models.UUIDField(default=uuid4, editable=False)  # legacy S3 key
```
A number of other platform models use "uuid" as the conventional name for this field.

submissions/models.py
```diff
@@ -693,3 +702,136 @@ def update_status(self, new_status):
...
+            >>> urls = file_manager.process_files(files)
+            >>> print(urls)
+            {'assignment.py': '/my_queue/550e8400-e29b-41d4-a716-446655440000'}
+        """
+        files_urls = {}
+        for filename, file_obj in files_dict.items():
+            # Validate file object type
+            if not (isinstance(file_obj, (bytes, ContentFile, SimpleUploadedFile)) or hasattr(file_obj, 'read')):
+                logger.error(f"Invalid file object type for {filename}")
+                raise InvalidFileTypeError(f"Invalid file object type for {filename}")
```
ormsbee, line 801: Error should say what the type actually was. I also don't think it's really necessary to create a new error for this, since this is exactly what `TypeError` is for.

submissions/models.py
```diff
@@ -693,3 +702,136 @@ def update_status(self, new_status):
...
+
+    class Meta:
+        indexes = [
+            models.Index(fields=['external_grader', 'uid']),
+        ]
+
+    @property
+    def xqueue_url(self):
+        """
+        Returns URL in xqueue format: /queue_name/uid
```
ormsbee, line 745: What does "xqueue format" mean?

submissions/models.py
```diff
@@ -693,3 +702,136 @@ def update_status(self, new_status):
...
+    file = models.FileField(
+        upload_to=submission_file_path,
+        max_length=512
+    )
+    original_filename = models.CharField(max_length=255)  # This is necessary to send file name to xqueue-watcher
+    created_at = models.DateTimeField(default=now)
+
+    class Meta:
+        indexes = [
+            models.Index(fields=['external_grader', 'uid']),
```
ormsbee, line 739: Do we ever look up by just uuid?

submissions/models.py
```diff
@@ -693,3 +702,136 @@ def update_status(self, new_status):
...
+            InvalidFileTypeError: If a file object has an unsupported type.
+            FileProcessingError: If there's an error reading from a file object,
+                                including I/O errors or Unicode decoding errors.
+
+        Example:
+            >>> file_manager = SubmissionFileManager(external_grader)
+            >>> files = {'assignment.py': file_obj}
+            >>> urls = file_manager.process_files(files)
+            >>> print(urls)
+            {'assignment.py': '/my_queue/550e8400-e29b-41d4-a716-446655440000'}
```
ormsbee, line 794: Is this a real example of something we'd expect to see in a capa problem code submission? If not, is it possible to make this example as realistic as possible?

ormsbee (changes requested):
Some comments and questions. I just want to make sure I'm understanding the intentions of this code properly. Thank you.

submissions/models.py
leoaulasneo98: Yes, it's a real example. I attach an image of the debug.
 image

submissions/models.py
leoaulasneo98: This is the format in which the xqueue-watcher receives the files. That is, the format in which the legacy XqueueServer delivers the files. I've corrected this to make it clearer.

submissions/models.py
```diff
@@ -693,3 +702,136 @@ def update_status(self, new_status):
...
+
+    @property
+    def xqueue_url(self):
+        """
+        Returns URL in xqueue format: /queue_name/uid
+        """
+        return f"/{self.external_grader.queue_name}/{self.uid}"
+
+
+class SubmissionFileManager:
```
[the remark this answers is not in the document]
leoaulasneo98: Hi Dave,
You're right about the naming confusion. Having a "xxxManager" class in models.py that isn't actually a Django ORM Manager is misleading and could cause confusion.

I've refactored this to follow a clearer naming convention:

Renamed SubmissionFileManager to SubmissionFileProcessor and move to api.py, which better describes its actual purpose - processing file operations for submissions rather than managing model queries.
Updated docstrings to reflect this change.
Kept the functionality encapsulated in a class since it maintains state (the external_grader reference) between operations.

This change preserves the existing functionality while making the code more intuitive.

submissions/models.py
leoaulasneo98: You're right on both counts. I've updated the code, thanks for the feedback.

submissions/models.py
```diff
@@ -693,3 +702,136 @@ def update_status(self, new_status):
...
+    def process_files(self, files_dict):
+        """
+        Process uploaded files from an Open edX environment and store them as SubmissionFile objects.
+
+        This method handles various file object types that might be received from Open edX, including:
+        - Native Open edX FileObjForWebobFiles objects
+        - Bytes objects
+        - ContentFile objects
+        - SimpleUploadedFile objects
+        - Any object with a 'read' method
```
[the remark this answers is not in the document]
leoaulasneo98: Dave, I understand your point. I initially designed this thinking of a scalable solution for various workflows, but I've reconsidered based on the actual needs of the project.
I've simplified the implementation by using standard TypeError instead of custom exceptions and adjusting the validation to focus on the essentials: bytes or objects with a 'read()' method.
Thanks for your feedback that leads us to more straightforward code.

submissions/models.py
leoaulasneo98: we don't currently look up by just uuid. The lookups are always through the external_grader relationship or using both fields. If we need direct uuid lookups in the future, we'll add a separate index for that field.

ormsbee (commented):
High level security question: Am I understanding correctly that this is storing things in the default Django storages backend, and if so, does that mean it will be publicly readable (though obscured because of the randomly generated UUIDs)?

leoaulasneo98:
> High level security question: Am I understanding correctly that this is storing things in the default Django storages backend, and if so, does that mean it will be publicly readable (though obscured because of the randomly generated UUIDs)?

Hi @ormsbee ,
The implemented flow uses Django file fields, which means security policies depend on the Bucket policies and Django configuration and integration. Here's a breakdown for the standard Tutor development setup using Minio:

You can access a resource if you have its complete URL.
You cannot access resources on ports 9000 or 9001 without credentials, except for resources with a complete, valid URL.
Even if you obtain a URL, you'll still need to decode it or know the file extension to access the resource. This is a mechanism inherited from XqueueServerLegacy.
Currently, there are no signature mechanisms to restrict client-only access. This is because the XqueueWatcher has broad capabilities, and the actual image processing is configured by the grader, not the XqueueWatcher itself.

If you wanted to implement a signature mechanism to restrict resource access, you would need to configure both the edX platform and the XqueueWatcher through the grader.
However, @angonz  and I have a question arising from your inquiry: the security for delivering the image depends on the bucket being used. Currently, they are stored in the default OpenEdX bucket, which means it adapts to its existing policies. But we want to know if it's appropriate to use this default bucket or if it would be more convenient to use another pre-existing bucket like the ORA-2 bucket (openedx-learning). Could you share your opinion on this point?

image

submissions/api.py
```diff
@@ -106,6 +112,101 @@ def create_external_grader_detail(student_item_dict,
...
+
+            # Handle file-like objects by reading their content
+            # This returns a bytes object that will be converted to ContentFile later
+            if hasattr(file_obj, 'read'):
+                try:
+                    file_obj = file_obj.read()  # read() returns a bytes object
+                except (IOError, OSError) as e:
+                    logger.error(f"Error reading file {filename}: {e}")
+                    raise FileProcessingError(f"Error reading file {filename}: {e}") from e
+                except UnicodeDecodeError as e:
```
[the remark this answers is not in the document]
ormsbee: I get that you're trying to be permissive here to make this friendlier for use in the platform, but please just require that the files here be instances of Django's [File](https://docs.djangoproject.com/en/5.2/ref/files/file/#the-file-object) object (including subclasses), and raise a `TypeError` if you find ones that are not. Let the caller worry about making sure the data exists there, and let it raise the exception when saving to `SubmisssionFile` if anything goes wrong. If someone has a `bytes` instance, it's easy enough for them to convert that to a `ContentFile` before calling this function. This "is it bytes or str or file" stuff is cluttering up the logic here without really giving much benefit, and bugs are more likely to slip in because of it.

ormsbee (changes requested):
Please also include a [version bump](https://github.com/openedx/edx-submissions/blob/master/submissions/__init__.py) as part of this PR. Thank you.

sarina:
@leoaulasneo98 could you please rebase and resolve conflicts?

leoaulasneo98:
> @leoaulasneo98 could you please rebase and resolve conflicts?

Hi Sarina, good afternoon. Ready.

submissions/models.py
```diff
@@ -693,3 +697,76 @@ def update_status(self, new_status):
...
+    and FileSystemStorage during tests.
+
+    This class automatically detects test environments and changes behavior accordingly.
+    """
+    def __new__(cls, *args, **kwargs):
+        if 'pytest' in sys.modules or any('test' in arg for arg in sys.argv):
+            return FileSystemStorage(*args, **kwargs)
+        return super().__new__(cls)  # pragma: no cover
+
+    bucket_name = getattr(settings, "FILE_UPLOAD_STORAGE_BUCKET_NAME", "openedx")
```
ormsbee, line 727: > However, @angonz and I have a question arising from your inquiry: the security for delivering the image depends on the bucket being used. Currently, they are stored in the default OpenEdX bucket, which means it adapts to its existing policies. But we want to know if it's appropriate to use this default bucket or if it would be more convenient to use another pre-existing bucket like the ORA-2 bucket (openedx-learning). Could you share your opinion on this point?

Sorry for not replying to this in my last review pass. Let's do this: have it look for a specific named storages defined like [how openedx-learning does it](https://github.com/openedx/openedx-learning/blob/655c0e350d3b275d19efa6393ec8822b4d944610/test_settings.py#L78C1-L86C1), but named `EDX_SUBMISSIONS['MEDIA']`. If that configuration value is not found (either `EDX_SUBMISSIONS` doesn't exist at all or `EDX_SUBMISSIONS['MEDIA']` doesn't), have it fall back to using the default Django storages. Please remember that you can set the `storage` parameter of a `FileField` to be any callable that returns a Storage class, so you can switch between storage configurations that way.

So please do the following:

1. Make a new function `get_storage()` and use the `functools.cache` decorator on it to cache it for future use.
2. Have `get_storage()` search for EDX_SUBMISSIONS configuration for the storage, but have it also gracefully fall back to the default storage backend if none is configured.
3. Make sure that `get_storage()` loads the storage configuration dynamically [like this](https://github.com/openedx/openedx-learning/blob/main/openedx_learning/apps/authoring/contents/models.py#L44-L47), and don't hardcode it to any specific storage type at this layer.
4. When creating the SubmissionFile.file FileField, set `storage=get_storage`.

As a general rule, try not to do smart switching behavior at this layer. Aside from making it hard to configure and override at the settings layer for testing, it's also just really confusing to have the `__new__` on a class return an instance of an entirely different class.

requirements/test.txt
```diff
@@ -1,42 +1,51 @@
...
     # via
-    #   -r /home/runner/work/edx-submissions/edx-submissions/requirements/docs.txt
+    #   -r /Users/leonardoberoes/Documents/Aulasneo/Projects/edx-submissions/requirements/docs.txt
     #   sphinx
 asgiref==3.8.1
     # via
-    #   -r /home/runner/work/edx-submissions/edx-submissions/requirements/base.txt
-    #   -r /home/runner/work/edx-submissions/edx-submissions/requirements/docs.txt
+    #   -r /Users/leonardoberoes/Documents/Aulasneo/Projects/edx-submissions/requirements/base.txt
+    #   -r /Users/leonardoberoes/Documents/Aulasneo/Projects/edx-submissions/requirements/docs.txt
```
ormsbee, line 18: These requirements shouldn't have your environment specific paths in them.

requirements/common_constraints.txt
```diff
@@ -3,6 +3,16 @@
...
+# This is a temporary solution to override the real common_constraints.txt
+# In edx-lint, until the pyjwt constraint in edx-lint has been removed.
+# See BOM-2721 for more details.
+# Below is the copied and edited version of common_constraints
+
+# This is a temporary solution to override the real common_constraints.txt
+# In edx-lint, until the pyjwt constraint in edx-lint has been removed.
+# See BOM-2721 for more details.
+# Below is the copied and edited version of common_constraints
+
```
ormsbee, line 15: It looks like these should just be deleted since they're copies of the first paragraph?

ormsbee (changes requested):
Thank you for addressing my concerns about bytes/file input handling.

Two blockers on this at the moment:

1. Storage configuration, which I've tried to give more explicit instructions about.
2. It looks like all the requirements files have been changed to have absolute paths from your local environment.

Would you like to schedule some time on Friday to go through any of this together?

Thank you.

requirements/test.txt
leoaulasneo98: Sorry, Dave. This was generated by compile requirements. It's also no longer necessary to add dependencies with the new approach.

submissions/models.py
leoaulasneo98: I've already applied the changes. It's made things much easier. Thanks for the help.

submissions/models.py
```diff
@@ -693,3 +696,99 @@ def update_status(self, new_status):
...
+
+    Returns:
+        Storage instance: Returns the configured storage if found in EDX_SUBMISSIONS['MEDIA'],
+                         otherwise returns Django's default_storage.
+
+    Example:
+        # In settings
+        from storages.backends.s3boto3 import S3Boto3Storage
+        EDX_SUBMISSIONS = {
+            'MEDIA': S3Boto3Storage(bucket_name='my-bucket')
```
ormsbee, line 744: Oh, I see where my misunderstanding of the code in `_get_storage_cached()` comes from. We shouldn't actually instantiate or even import the storage class in the settings files, only specify the `EDX_SUBMISSIONS['MEDIA']` configuration dict in the standard convention that Django uses for media config, which specifies the path of the class we'll need to load later as a string.

The reason has to do with when these things get loaded in the startups cycle. Settings are often overridden in multiple places (e.g. `common.py`, `prod.py`, `tutor.py`, etc). But there's a certain amount of caching/bootstrapping that happens when these storage backends get imported and initialized. For example, there was this issue:

https://github.com/overhangio/tutor-minio/pull/25

> The value for AWS_S3_ENDPOINT_URL set in openedx-development-settings specifies port 9000, and it was being written to Studio and LMS settings correctly. But this is the sequence of what was really happening:
> 
> 1. openedx-common-settings first sets AWS_S3_ENDPOINT_URL WITHOUT port
>    info, since this is how it's used in prod deployments.
> 2. openedx-common-settings imports S3Boto3Storage because it wants to
>    make a subclass to use for USER_TASKS_ARTIFACT_STORAGE.
> 3. The act of importing that package has the side-effect of initializing
>    the Django storages (django.core.files.storage.default_storage).
> 4. Because the DEFAULT_FILE_STORAGE class is S3Boto3Storage, it then
>    reads the value of AWS_S3_ENDPOINT_URL (which has no port info), and
>    sets this as the S3 conneciton information for the S3Boto3Storage
>    instance at that time.
> 5. openedx-development-settings then comes along and changes the value
>    of AWS_S3_ENDPOINT_URL to specify port 9000, but by this point it's
>    too late to affect django-storages (though it will confuse folks who
>    try to debug).

Leaving it in string form avoids this problem. The settings can get overridden as much as people need, and the storage class won't be instantiated until after it's all been set up properly. Please assume the configuration will be something like:

```python
EDX_SUBMISSIONS = {
    'MEDIA': {
        'BACKEND': 'django.core.files.storage.InMemoryStorage',
        'OPTIONS': {
        }
    }
}
```

Also, please try specifying such a backend for the [test settings](https://github.com/openedx/edx-submissions/blob/master/settings.py), to make sure it works properly.

ormsbee (changes requested):
In addition to the code comments below, please also dothe following:

1. Update [the version](https://github.com/openedx/edx-submissions/blob/master/submissions/__init__.py#L2).
2. When you push your changes, please don't squash your commits. We can squash at the end once everything is approved, but it makes it easier/faster to review I can just "view changes since last review", and that functionality breaks if you squash.

Thank you.

submissions/errors.py
```diff
@@ -40,6 +40,17 @@ class ExternalGraderQueueEmptyError(SubmissionError):
...
 
+class FileProcessingError(SubmissionError):
+    """
+    Exception raised when there's an error reading or processing a file.
+
+    This exception is raised when file operations fail, such as:
+    - I/O errors when reading file content
+    - OS errors during file operations
+    - Unicode decoding errors when processing file content
+    """
```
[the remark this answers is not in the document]
leoaulasneo98: Ok Dave, understood

submissions/models.py
```diff
@@ -693,3 +696,99 @@ def update_status(self, new_status):
...
+def _get_storage_cached():
+    """
+    Cached implementation to get the configured storage backend.
+    This function is for internal use only.
+    """
+    edx_submissions_config = getattr(settings, 'EDX_SUBMISSIONS', {})
+    storage_config = edx_submissions_config.get('MEDIA')
+
+    if storage_config:
+        return storage_config
```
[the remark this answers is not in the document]
leoaulasneo98: Hi Dave, I had the same question while reading your previous instructions and preferred to keep things simple. But I completely understand what you're saying, and with that piece of code you've made my job much easier. I had previously tried instantiating the storage class, but the result wasn't as clean as I would have liked, so I went with the solution you reviewed a few days ago.

submissions/models.py
leoaulasneo98: Understood Dave, I'll try to have the changes made before the meeting.

ormsbee (approved):

submissions/api.py
```diff
@@ -15,7 +15,6 @@
 # SubmissionError imported so that code importing this api has access
 from submissions.errors import (  # pylint: disable=unused-import
     ExternalGraderQueueEmptyError,
-    SubmissionError,
```
timmc-edx, line 18: This removal appears to be causing a failure in edx-platform when upgrading to edx-submissions 3.11:

https://github.com/openedx/edx-platform/actions/runs/15213754304/job/42794120855?pr=36786

ormsbee: @timmc-edx: Thank you! @leoaulasneo98: Can you please fix this and do a patch version bump?

leoaulasneo98: Hi Dave, I'll take care of it. Sorry for the mistake. I'll add the import again and do the corresponding PR.

ormsbee: I merged https://github.com/openedx/edx-submissions/pull/301 and made a 3.11.1 release.

PR #287, opened 2025-02-24
FC-73 Add XqueueViewset with xwatcher services (Feature: XQueueViewSet - Compatible API Endpoints for External Grader Integration)

leoaulasneo98:
# FC-73 Feature: XQueueViewSet - Compatible API Endpoints for External Grader Integration

⚠️ **Important:** This PR builds on the `ExternalGraderDetail` and `SubmissionFile` infrastructure deployed in previous PRs (FC-73). Please ensure those changes are fully deployed before merging.

## Description

This pull request implements the XQueueViewSet, providing compatible API endpoints that allow external graders (XWatcher) to interact with the new submission and grading architecture. This viewset consolidates authentication, submission retrieval, and result processing services while leveraging the new `ExternalGraderDetail` and `SubmissionFile` models.

## Motivation

The current XQueue implementation requires a strategic update to:
* Provide a secure and efficient API for external grader integration
* Simplify the complex multi-system communication patterns
* Enhance session management and security
* Integrate cleanly with the new submission architecture

### Authentication Services
* Custom `XQueueSessionAuthentication` class with:
  * CSRF exemptions for result endpoints
  * Robust session management with explicit cookie handling
  * Secure login/logout endpoints

### Submission Distribution
* Queue-based submission retrieval through `get_submission` endpoint
* Integrated file handling through `SubmissionFileManager`
* Status tracking with explicit state transitions
* UUID-based submission keys for enhanced security

### Result Processing
* Transactional score updates via the `put_result` endpoint
* Comprehensive error handling with detailed logging
* Atomic status updates for reliable state management
* Automatic retry mechanism for failed submissions

### Session Management
* Enhanced session parsing and validation
* Explicit cookie extraction from request headers
* Secure session key generation and validation

### Score Processing
* Integration with the submissions API via `set_score`
* Robust failure tracking and processing
* Status transition enforcement

### Error Handling
* Comprehensive validation of grader responses
* Detailed logging for troubleshooting
* Clear error responses for client integration

## Testing Strategy

Comprehensive test coverage includes:
* Authentication workflow verification
* Submission retrieval process
* Result processing and error handling
* Session management edge cases
* Integration with existing submission models

## Open edX Compliance

This implementation adheres to Open edX standards through:
* RESTful API design following DRF best practices
* Comprehensive security measures
* Transaction safety for critical operations
* Extensive logging for operational visibility
* Compatibility with existing systems

## Performance Considerations

* Optimized database queries with select_related
* Efficient transaction handling
* Minimal processing overhead

**BREAKING CHANGES**: None. Designed for full backward compatibility with existing XWatcher services.

## Documentation

Updated documentation will include:
* XQueue API specification for external graders
* Session management guidelines
* Error handling and response format specifications
* Integration guide for external grader services

## Implementation References

* Related ADR: Implementation of XQueue Compatible Views for External Grader Integration
* Previous PRs: FC-73 SubmissionQueueRecord and SubmissionFile implementations


leoaulasneo98:
@ormsbee Hi Dave, we have uploaded two new Prs to draft because they are still being worked on, and we want to ask a question in this PR. We were implementing tests to prevent future errors as you mentioned the other day in the meeting. Basically we add the atomic decorator to get submission to ensure that no root conditions exist. To verify that this was the case, raise a multi-threaded instance through the local run tutor (as if it were a productive environment). When running this I detected an error that does not happen in the development environment. Basically xwatcher is not sending the request.user through the request, for this our solution was to add it before calling the dispatch method. This in development works (however it is a far-fetched solution in my opinion). But when we run the productive environment, the edx paltform SafeSecureSession Middleware triggers an exception. The problem is the disparity in authentication methods between xwatcher and edx-platform. Since we cannot manipulate xwatcher or the base configuration of edx platform, our range of action is quite limited. We wanted to ask you if you or anyone on the team has any ideas regarding this scenario. Personally I think we should start considering adding an optional auth flow in xwatcher, but if you have a better idea it's welcome. I look forward to your comments.

angonz:
Review can start only after #283 and #286 are merged

ormsbee:
> Hello @ormsbee , good afternoon, greetings. We have a question about this here, we already have the put result service in edx submission (which also exists in Xqueue server). We are missing implementing the call back but this has a drawback, in edx submission we cannot import xqueue_callback or any open edx package because there start to be import errors when running the tests, in addition to the fact that a circular import is created since edx submission is a dependency of edx platform.
> 
> It is for this reason that we decided to use an open edx event signal so that the callback is executed once the score of the submission has been stored. However, we have some questions:
> 
> 1 Is this the best solution at the architectural level

I think it's totally reasonable, and one of the best uses of signals.

> 2 Are there possible errors or incompatibilities that we are overlooking

Nothing comes to mind. We have to be careful about any change to the data being sent in the signal, but I can't think of any other issues.

> 3 Should we make a new callback or recycle the xqueue callback

I think the first thing I would try is to make a new function that listens to the signal and translates the data into whatever the existing xqueue callback function expects. But please feel free to refactor differently if you find it's easier to do otherwise.

> 4 Is there a more efficient solution?

I wouldn't worry too much about efficiency with this. As long as the callback is being invoked in-process, I don't think it'll make any difference to use a signal vs. other mechanisms.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+                    'xqueue_body': json.dumps(submission_data),
+                    'xqueue_files': json.dumps(get_files_for_grader(external_grader))
+                }
+
+                return Response(
+                    self.compose_reply(True, content=json.dumps(payload)),
+                    status=status.HTTP_200_OK
+                )
+
+            except ValueError as e:
```
ormsbee, line 170: Where is this `ValueError` coming from?

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class XqueueViewSet(viewsets.ViewSet):
+    """
+    A collection of services for xqueue-watcher interactions and authentication.
+
+    This ViewSet provides endpoints for managing external grader results
+    and handling user authentication in the system.
```
ormsbee, line 30: Please add a note in the comments about how this is re-building an API-compatible implementation for xqueue-watcher to query, which helps to explain some of the non-standard choices we're making. Also link the DEPR ticket.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+            try:
+                if external_grader.status != "pulled":
+                    external_grader.update_status("pulled")
+
+                submission_data = {
+                    "grader_payload": json.dumps({"grader": external_grader.grader_file_name}),
+                    "student_info": json.dumps({
+                        "anonymous_student_id": str(external_grader.submission.student_item.student_id),
+                        "submission_time": str(int(external_grader.created_at.timestamp())),
+                        "random_seed": 1
```
ormsbee, line 151: Why is `random_seed` hardcoded?

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+                self.compose_reply(False, "Submission already in process"),
+                status=status.HTTP_409_CONFLICT
+            )
+
+        if external_grader:
+            try:
+                if external_grader.status != "pulled":
+                    external_grader.update_status("pulled")
+
+                submission_data = {
```
ormsbee, line 146: Where should I look in the codebase to see how this data is created for legacy XQueue submissions today?

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+                    "student_response": external_grader.submission.answer
+                }
+
+                payload = {
+                    'xqueue_header': json.dumps({
+                        'submission_id': external_grader.submission.id,
+                        'submission_key': external_grader.pullkey
+                    }),
+                    'xqueue_body': json.dumps(submission_data),
+                    'xqueue_files': json.dumps(get_files_for_grader(external_grader))
```
ormsbee, line 162: Doesn't this double escape the `xqueue_files` so that the whole field (list?) comes over as a string--so the client will have to parse the JSON for the response as a whole and then json.loads on the string value for `xqueue_files` to get the list? Is that actually what happens today? If so, please note it in a comment, because that is really counter-intuitive.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+                status=status.HTTP_400_BAD_REQUEST
+            )
+
+        try:
+            external_grader = ExternalGraderDetail.objects.select_for_update(
+                                                                        nowait=True).get(submission__id=submission_id)
+        except ExternalGraderDetail.DoesNotExist:
+            log.error(
+                "Grader submission_id refers to nonexistent entry in Submission DB: "
+                "grader: %s, submission_key: %s, score_msg: %s",
```
ormsbee, line 204: This log entry says "grader" but the field being written is "submission_id". Also, please keep the old terminology of "grader_reply" rather than "score_msg", since it contains more than just the score, and it's not just a text message.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+                score_msg
+            )
+            return Response(
+                self.compose_reply(False, 'Submission does not exist'),
+                status=status.HTTP_404_NOT_FOUND
+            )
+
+        if not external_grader.pullkey or submission_key != external_grader.pullkey:
+            log.error(f"Invalid pullkey: submission key from xwatcher {submission_key} "
+                      f"and submission key stored {external_grader.pullkey} are different")
```
ormsbee, line 216: Please log the key they tried to send for this entry, since it might aid debugging, e.g. if they're only different by case, or if they're different by whitespace, etc.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+            return fail
+
+        try:
+            header = external_reply['xqueue_header']
+            score_msg = external_reply['xqueue_body']
+        except KeyError:
+            return fail
+
+        try:
+            header_dict = json.loads(header)
```
ormsbee, line 256: It looks like the code in this PR is unnecessarily encoding and decoding JSON internally, and I'm not clear on why it's doing so. We should keep things in native data structures as much as possible internally, and only keep things as JSON encoded if we're doing nothing with it but passing it through (i.e. we're not inspecting it), or if backwards compatibility requires that we do it that way.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+            return fail
+
+        if not isinstance(header_dict, dict):
+            return fail
+
+        for tag in ['submission_id', 'submission_key']:
+            if tag not in header_dict:
+                return fail
+
+        submission_id = int(header_dict['submission_id'])
```
ormsbee, line 273: Why would this ever not already be an int?

submissions/views/xqueue.py
ormsbee: Where we're dealing with JSON encoded messages coming _in_, we should parse it to native data structures as soon as it's practical to do so.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+        try:
+            header_dict = json.loads(header)
+        except (TypeError, ValueError):
+            return fail
+
+        try:
+            score = json.loads(score_msg)
+            points_earned = score.get("score")
+        except (TypeError, ValueError):
+            return fail
```
ormsbee, line 264: Adding error logging to this validation would be nice, since the only thing people running this in production will currently see is that the replies are bad, but not why they're bad.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+    def login(self, request):
+        """
+        Endpoint for authenticating users and creating sessions.
+        """
+        log.info(f"Login attempt with data: {request.data}")
+
+        if 'username' not in request.data or 'password' not in request.data:
+            return Response(
+                {'return_code': 1, 'content': 'Insufficient login info'},
+                status=status.HTTP_400_BAD_REQUEST
```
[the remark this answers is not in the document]
ormsbee: Actually, given how many places the return codes are changing, it's probably better to add a comment in the docstring for the viewset addressing this.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+                .select_for_update(nowait=True)
+                .filter(
+                    Q(queue_name=queue_name, status='pending') |
+                    Q(queue_name=queue_name, status='pulled', status_time__lt=timeout_threshold)
+                )
+                .select_related('submission')
+                .order_by('status_time')
+                .first()
+            )
+        except DatabaseError:
```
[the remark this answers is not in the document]
leoaulasneo98: We removed the try/except block for DatabaseError in get_submission because, after implementing skip_locked (as discussed in our meeting), locking conflicts are now avoided and this exception is no longer expected in this context.

submissions/views/xqueue.py
```diff
@@ -0,0 +1,296 @@
...
+        - queue_name (required): Name of the queue to pull from
+        Returns:
+        - Submission data with pull information if successful
+        - Error message if queue is empty or invalid
+        """
+        queue_name = request.query_params.get('queue_name')
+
+        if not queue_name:
+            return Response(
+                self.compose_reply(False, "'get_submission' must provide parameter 'queue_name'"),
```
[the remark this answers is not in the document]
leoaulasneo98: That's correct—the queue name is not validated for existence or correctness at this point. The queue name comes from the exercise's XML configuration and is defined by the course author. We rely on the value provided in the XML, so it's up to the author to ensure it's valid and consistent with the grader setup.
image

submissions/views/xqueue.py
leoaulasneo98: Legacy XQueue stores the structured xqueue_body as-is from the LMS (in submission_queue/ext_interface.py). The new system reconstructs this exact structure to maintain API compatibility.
The original structure is created in the LMS before being sent to XQueue - legacy just stores/returns it, while the new system recreates it from normalized data.

image

image


image

leoaulasneo98: First image it is a comparison between edx submissions and xqueue server response (the same response basically).  Seconde image is xqueue server code and Third image is the xqueue server response in debug console.

submissions/views/xqueue.py
leoaulasneo98: Yes, you're absolutely right about the double escaping.
Looking at the legacy response data, this counter-intuitive behavior does exist today:

image

The legacy system works this way, so we're maintaining the same (admittedly confusing) API contract.

xqueue_files comes as string '{}' not object {}
Client must: json.loads(response) → json.loads(content) → json.loads(xqueue_files)

Below I present images of the xqueue server:
image

image

image

submissions/views/xqueue.py
leoaulasneo98: You're right that this is counter-intuitive, but we're constrained by legacy compatibility.
The legacy XQueue server returns exactly this nested JSON string format:

{
  "xqueue_body": "{\"grader_payload\": \"{...}\", \"student_info\": \"{...}\", \"student_response\": \"...\"}"
}

xqueue-watcher clients expect and parse this format:

json.loads(response['content']) → gets payload
json.loads(payload['xqueue_body']) → gets submission data
json.loads(submission_data['grader_payload']) → gets grader config

We have to replicate this exact behavior - the double/triple encoding is required for API compatibility with existing xqueue-watcher deployments.
The encoding/decoding dance exists because that's how the legacy system works and what clients expect.

This is xqueue server response
image

submissions/views/xqueue.py
leoaulasneo98: You're technically correct that xqueue-watcher sends it as an int.
However, this int() conversion exists in the legacy code for defensive programming - JSON parsing can sometimes deserialize numbers as strings depending on the client implementation or how the data was originally encoded.
Legacy XQueue does the same conversion:

image

We're maintaining this pattern for:

Legacy compatibility - same behavior as original
Robustness - handles edge cases where JSON might come as string "22" instead of int 22

The conversion is essentially defensive programming that the legacy system used, so we're keeping it consistent.

submissions/views/xqueue.py
leoaulasneo98: The random_seed was OMITTED in PR https://github.com/openedx/edx-platform/pull/36258 where LMS→submissions integration was added.

Proposed fix:
1. Add random_seed field to ExternalGraderDetail with default=1 in PR #287
    * When LMS doesn't have random_seed defined in the XML (like your test exercise), LMS sends 1 by default

2. Include random_seed parameter in LMS integration in PR https://github.com/openedx/edx-platform/pull/36408

The hardcoded 1 correctly matches legacy behavior for cases where no random_seed is specified in the problem XML.

image

leoaulasneo98:
@ormsbee 
General comment:
We tested this flow with two types of exercises: Python scripts and file submissions.
For the xqueue-watcher, we implemented a simple grader to validate Python code, while file submissions depend on the specific grader implementation - there's no standard defined in xqueue-watcher for file handling.
I invite you to check out our demo video on YouTube (private, accessible only with the link). The code shown is somewhat outdated, but the functionality remains the same - the changes we've made improve efficiency and versatility but don't change the workflow.
The video provides reliable proof that what we implemented in this view works end-to-end with xqueue-watcher, especially addressing your concerns about the response format compatibility.
The video demonstrates that our API maintains full compatibility with existing xqueue-watcher deployments.

https://www.youtube.com/watch?v=Ac6wa_aiLXw

image

image
image

image

leoaulasneo98:
@ormsbee The login testing features are implemented as you told me in the last meeting

PR #303, opened 2025-05-27
fix: Restore missing error export

timmc-edx:
This was removed in 3.11.0 but should remain re-exported for edx-platform and any other users.


timmc-edx:
Ah, thanks. Didn't see the other one!

PR #352, opened 2026-05-26
fix: obtain CSRF cookie via GET /xqueue/login/ for xqueue-watcher session auth

blarghmatey:
## Summary

The `XQueueViewSet` was using DRF's standard `SessionAuthentication`, which enforces CSRF validation on all mutating requests from authenticated users. This caused `POST /xqueue/put_result/` to return a 403 when called by `xqueue-watcher`.

### Root cause

`xqueue-watcher` authenticates via `POST /xqueue/login/` then uses the resulting session cookie for subsequent calls. Before the login POST it does a pre-fetch GET to the same URL to obtain a CSRF cookie — but the login action is declared `methods=['post']` only, so the GET returns 405 with no cookie set. The watcher therefore has no CSRF token to send, and when it POSTs to `put_result/` with a valid session, DRF's `SessionAuthentication.enforce_csrf()` fails the check and returns 403.

`get_submission` succeeds because it is a GET request, which is CSRF-exempt by design.

### Fix

Extract `CsrfExemptSessionAuthentication` into `submissions/authentication.py` and use it in `XQueueViewSet`. CSRF is not appropriate for this service-to-service API; authorization is already enforced by the `IsXQueueUser` permission class, which requires the authenticated user to be a member of the `xqueue` group.

The test module previously defined its own local copy of this class for test setup — it now imports from the production module instead.

## Testing

- All 23 existing viewset tests pass
- `isort` and `pycodestyle` clean
- Note: the `pylint` `django-not-configured` error on `submissions/__init__.py` is a pre-existing failure on `master` unrelated to this change


openedx-webhooks:
Thanks for the pull request, @blarghmatey!


This repository is currently maintained by `@openedx/committers-edx-submissions`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>


---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

ormsbee (commented):
> Before the login POST it does a pre-fetch GET to the same URL to obtain a CSRF cookie — but the login action is declared methods=['post'] only, so the GET returns 405 with no cookie set. The watcher therefore has no CSRF token to send, and when it POSTs to put_result/ with a valid session, DRF's SessionAuthentication.enforce_csrf() fails the check and returns 403.

Could this be solved by having the CSRF cookie sent in the login GET then? I am a bit nervous about the idea that we are relying on xqueue users never overlapping with real users.

blarghmatey:
Thanks for the feedback @ormsbee. You're right — the cleaner fix is to support the CSRF flow that xqueue-watcher already implements rather than bypassing CSRF entirely.

**Root cause recap:** xqueue-watcher GETs `/xqueue/login/` before POSTing credentials specifically to obtain a CSRF cookie. Since the action only accepted `POST`, the GET returned 405 and no cookie was set.

**Updated approach (latest commit):**
- Add `GET` to the login action. The handler calls `get_token(request)` which ensures Django sets the `csrftoken` cookie in the response.
- Revert `authentication_classes` back to the standard `SessionAuthentication` — no CSRF bypass needed.
- Delete `submissions/authentication.py` since `CsrfExemptSessionAuthentication` is no longer used.

**Why this works end-to-end:** xqueue-watcher's `_login()` already does exactly this:
1. GETs `/xqueue/login/` → reads `csrftoken` cookie from response
2. POSTs credentials with `X-CSRFToken` header (CSRF is not enforced here since the request is unauthenticated)
3. After login, stores the CSRF token in `self.session.headers` so every subsequent `put_result` POST automatically includes `X-CSRFToken`

All 19 xqueue-watcher client tests pass unchanged, including the ones that explicitly verify the GET→cookie→POST CSRF flow.

The updated edx-submissions tests replace `test_put_result_succeeds_without_csrf_token` (which tested the bypass) with:
- `test_login_get_sets_csrf_cookie` — verifies GET returns 200 with `csrftoken` cookie set
- `test_xqueue_watcher_csrf_flow` — end-to-end test with `enforce_csrf_checks=True` proving no CSRF exemption is needed

blarghmatey:
## Production investigation findings

Following up with details from a live production investigation (mitxonline, Django 5.2.15 + DRF 3.x, edx-submissions 4.0.1).

### What we observed

`xqwatcher` was returning 403 on every `put_result` POST immediately after pod startup.  The `get_submission` GET succeeded, the grader ran, and then every `put_result` 403'd.  After one re-auth attempt the watcher gave up and the submission was stuck.

### Root cause — two separate failures stacked

**Failure 1 (addressed by this PR):** `_login()` was posting credentials without first GETting the login page.  No CSRF cookie → no `X-CSRFToken` header on `put_result` → DRF's `SessionAuthentication.enforce_csrf()` rejected it.

This PR (GET login endpoint) directly fixes failure 1.

**Failure 2 (Django 4+ / HTTPS-only):** Even with a valid CSRF token, Django's `CsrfViewMiddleware.process_view()` applies an additional check on HTTPS POST requests.  If neither `HTTP_ORIGIN` nor `HTTP_REFERER` is present it raises `REASON_NO_REFERER` _before_ checking the token:

```python
# django/middleware/csrf.py  (Django 4+)
def process_view(self, request, callback, ...):
    ...
    if request.is_secure():
        self._check_referer(request)   # raises RejectRequest if Referer is None
    self._check_token(request)
```

`requests.Session` does not add `Referer` automatically, so xqwatcher's `put_result` POST always hit this path.  The 403 was logged by `django.request` (not `django.security.csrf`), which is why it looked like a permissions failure rather than a CSRF failure.

### How it manifests in logs

The 403 appears in LMS logs as:

```
django.request WARNING Forbidden: /xqueue/put_result/
```

There is **no** corresponding `django.security.csrf` entry because `DRF SessionAuthentication.enforce_csrf()` raises `PermissionDenied` directly rather than returning a `HttpResponseForbidden` via the CSRF middleware.

### Fix on the xqwatcher side

openedx/xqueue-watcher#125 addresses both failures:

1. GET `/xqueue/login/` before POSTing (uses this PR's new endpoint)
2. After login, persist `Referer: <xqueue_server>` in `session.headers` so every subsequent POST passes Django's `_check_referer()` automatically

The two fixes together produce a clean sequence:

```
GET  /xqueue/login/       → 200 + csrftoken cookie
POST /xqueue/login/       → 200 (Referer + X-CSRFToken set for this request)
# session headers: {Referer: <server>, X-CSRFToken: <token>}
POST /xqueue/put_result/  → 200 ✓
```

### Recommendation for this PR

The GET login endpoint (this PR) is the necessary server-side piece.  Without it, `_login()` cannot obtain a CSRF cookie even after the xqwatcher-side fix lands, so older edx-submissions deployments would still fail.

The `CsrfExemptSessionAuthentication` approach from the original commit would also work, but @ormsbee's suggestion (and the updated approach in this PR) is cleaner — it preserves CSRF enforcement while giving the client a proper way to obtain a token.

ormsbee (approved):
Note to anyone glancing at this PR: This PR is also doing code re-formatting for the files that are touched, and so the diff looks bigger than it is.

